PR #8, opened 2021-01-05
Dev

abinashkonersman:
Built translator from CDOT data and iCone data feed to WZDx 3.0 
Created unit tests with 90 plus percent of code coverage
Added  documents directory with icone and wzdx specification related documents
Added sample files


TonyEnglish (approved):
Looks good, thx!

PR #12, opened 2021-01-19
Dev

abinashkonersman:
Integrated iCone sensor data into WZDx description
Implemented translator in Google Cloud Function
Increased unit test coverage


README.md
```diff
@@ -34,7 +41,28 @@ This code requires Python 3.6 or a higher version. If you haven’t already, dow
...
+```
+python icone_translator.py -i '../sample files/icone data/incidents_extended.xml' 
+```
+
+### Testing
+
+#### Step 1: Run the unit test for translator script (from root directory)
+
+```
+coverage run -m pytest 'test/' -v
```
payneBrandon: Running this doesn't work out of the box. What is "coverage"? Is this an additional tool required to perform the testing? If so we need to mention that here.

README.md
```diff
@@ -34,7 +41,28 @@ This code requires Python 3.6 or a higher version. If you haven’t already, dow
...
+
+```
+coverage run -m pytest 'test/' -v 
+```
+### Google Cloud Function
+
+A system was created in google cloud platform to automatically translate iCone data to WZDx message.This system consits of two buckets and a cloud function.
+When a file is uploaded to the first bucket, the cloud function automatically downloads the file, translates to a WZDx message and uploads it to the second bucket.
+
+![alt text](https://github.com/CDOT-CV/Work_Zone/blob/dev/translator/GCP_cloud_function/iCone%20Translator%20block%20diagram.png)
```
payneBrandon: We'll want this to be a relative link (read [here](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-readmes#relative-links-and-image-paths-in-readme-files) ) rather than pointing to a specific branch.

README.md
```diff
@@ -34,7 +41,28 @@ This code requires Python 3.6 or a higher version. If you haven’t already, dow
...
+
+### Documentation
 
+documentation for iCone to WZDx translator is located here: https://github.com/CDOT-CV/Work_Zone/tree/main/translator/docs
```
payneBrandon: As above, this should  be a relative link so we avoid issues pointing to a specific branch that may or may not have documents.

translator/GCP_cloud_function/README.md
```diff
@@ -0,0 +1,18 @@
+# iCone translator Cloud Function
+
+A system was created in google cloud platform to automatically translate iCone data to WZDx message.This system consits of two buckets and a cloud function.
+When a file is uploaded to the first bucket, the cloud function automatically downloads the file, translates to a WZDx message and uploads it to the second bucket.
+
+![alt text](https://github.com/CDOT-CV/Work_Zone/blob/dev/translator/GCP_cloud_function/iCone%20Translator%20block%20diagram.png)
```
payneBrandon: Again, relative links needed

translator/GCP_cloud_function/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+## files present
+cloud_function.zip
```
payneBrandon: Instead of a zip file, let's add these files in uncompressed so we can review them and push changes as needed.

payneBrandon:
The `iCone Translator block diagram.png` has a spelling error for "translates" that will need to be addressed.

translator/source_code/icone_translator.py
```diff
@@ -37,11 +38,25 @@ def wzdx_creator(messages, info):
     if info['metadata'].get('datafeed_frequency_update', False):
         data_source['update_frequency'] = info['metadata']['datafeed_frequency_update']
     data_source['update_date'] = datetime.now().strftime("%Y-%m-%dT%H:%M:%SZ")
-    data_source['location_verify_method'] = info['metadata']['location_verify_method']
+    # data_source['location_verify_method'] = info['metadata']['location_verify_method']
```
payneBrandon: There's a large amount of commented code in here that doesn't seem to be needed. Can we clean that up and remove?

translator/source_code/icone_translator.py
```diff
@@ -20,7 +21,7 @@ def wzdx_creator(messages, info):
     # hardcode
     wzd['road_event_feed_info']['feed_info_id'] = info['feed_info_id']
     wzd['road_event_feed_info']['update_date'] = datetime.now().strftime("%Y-%m-%dT%H:%M:%SZ")
-    wzd['road_event_feed_info']['publisher'] = 'CDOT '
+    wzd['road_event_feed_info']['publisher'] = 'CDOT'
```
payneBrandon, line 19: Good catch!

translator/source_code/icone_translator.py
```diff
@@ -65,9 +80,13 @@ def wzdx_creator(messages, info):
         # print("")
         # print("")
         # Parse Incident to WZDx Feature
-        wzd['features'].append(parse_incident(incident))
-
+        polyline = parse_polyline(incident['location']['polyline'])
+        avg_point = [(polyline[0][0] + polyline[-1][0]) / 2, (polyline[0][1] + polyline[-1][1]) / 2]
 
+        if is_in_colorado(avg_point) or True:
```
payneBrandon: What are we doing with this check? It looks like we're checking if the average point on the polyline is in Colorado, then proceeding regardless. If that's the case, do we even need this check? Or the `avg_point` calculation for that matter?

translator/source_code/icone_translator.py
```diff
@@ -87,25 +106,38 @@ def wzdx_creator(messages, info):
...
+
+
+def is_in_colorado(coordinates):
+    point = Point(coordinates[0], coordinates[1])
+    colorado = Polygon(
+        [(36.993016, -102.042089), (36.999084, -109.045223), (41.000659, -109.050076), (41.002361, -102.051721)])
+    return polygon_contains_point(point, colorado)
+
+
+is_in_colorado([40.703547, -102.584596])
```
payneBrandon: Was this for debugging? Can we remove?

translator/source_code/icone_translator.py
```diff
@@ -192,11 +322,19 @@ def parse_incident(incident):
...
+    properties['road_name'] = incident['location'].get('street', '')
 
     # direction
-    direction = ''
-    properties['direction'] = get_road_direction(geometry['coordinates'])
+    direction = parse_direction_from_street_name(incident['location'].get('street', ''))
+    if not direction:
+        direction = get_road_direction(geometry['coordinates'])
+        if not direction:
+            return None
```
payneBrandon, line 290: Is this filtering out all single point messages? IE if an iCone message has two points and they are the same location, would that get filtered out here with this `return None`? I think those messages would still be valid.

translator/source_code/icone_translator.py
```diff
@@ -136,29 +190,105 @@ def get_event_status(start_time_string,end_time_string):
...
+    if incident.get('sensor'):
+        description += '\n sensors: '
+        for sensor in incident['sensor']:
+            if not isinstance(sensor, str):
+                if sensor['@type'] == 'iCone':
+                    description += '\n' + json.dumps(parse_icone_sensor(sensor), indent=2)
+            else:
+                sensor = incident['sensor']
+                if sensor['@type'] == 'iCone':
+                    description += '\n' + json.dumps(parse_icone_sensor(sensor), indent=2)
```
payneBrandon, line 160: What case is this handling? The sensor appears to always be an object and not a string so the check on 199 should always go through right? Am I missing a case?

translator/source_code/icone_translator.py
```diff
@@ -307,62 +438,67 @@ def add_ids(message, add_ids):
...
+
+inputfile, outputfile = parse_arguments(sys.argv[1:])
+
+
+def initialize_info():
     info = {}
 
     #### Consider whether this id needs to be hardcoded or generated
-    info['feed_info_id'] = "feed_info_id"
+    info['feed_info_id'] = "104d7746-688c-44ed-b195-2ee948bf9dfa"
```
payneBrandon, line 404: Where is this GUID coming from?

payneBrandon (changes requested):
A few comments for you @abinashkonersman. Looking good!

README.md
abinashkonersman: i changed it ..can you please take a look and let me know if thats ok or not.Thanks

abinashkonersman:
> The `iCone Translator block diagram.png` has a spelling error for "translates" that will need to be addressed.

corrected the spelling

translator/GCP_cloud_function/README.md
abinashkonersman: updated to relative link

README.md
abinashkonersman: updated to relative link

translator/source_code/icone_translator.py
abinashkonersman: We will remove this and it was there before we knew that iCone was pre-filtering the data for Colorado.

translator/source_code/icone_translator.py
abinashkonersman: We will remove this as well.

translator/source_code/icone_translator.py
abinashkonersman: Yes, we are ignoring incidents with no direction or road name information.Both of the fields are required in the WZDx message and if we can not determine the road name or the road direction (N,E,S,W), then a WZDx message cannot be created for that incident.

translator/source_code/icone_translator.py
jacob6838: Short answer: this is for a very specific but required case

Long answer: We are using the xmltodict library to read xml into a dictionary object in python. In xml, you can have multiple tags with the same name, and this occurs a lot in the incidents_extended.xml iCone data. Dictionaries must have unique keys, so when this happens the xmltodict library imports these objects as a list of objects. We decided to use a for loop to iterate through this list of objects. Inconveniently, when there is only a single key in the xml, the xmltodict library takes it in as a dictionary and not a list, and then the for loop iterates over the dictionary, reading invididual keys and values, not full dictionary objects like we were expecting. To combat this, we check if the iterated object is a string. If it is, then we access its' parent as the object and do the same thing we would have done.

Hopefully that made sense

translator/source_code/icone_translator.py
jacob6838: This was a randomly generated UUID. The WZDx feed_info_id should remain the same each time a feed is updated, to show that we are updating a previous data feed and not creating an entirely new one. We assumed that we would always be updating the same feed, since we will always be drawing from the same data source. 
If this is not the case, then we can alter this logic to randomly generate a new UUID, but then we would always be generating a new WZDx feed.

payneBrandon: Ah I see, that makes sense now. Thanks for the clarification! I think for now at least we plan to continue consuming from the same data feed and can leave this id alone.

translator/source_code/icone_translator.py
jacob6838: What Abinash said. We added this as an attempt to filter the data for just Colorado. The incidents_extended data we have been working with is not filtered by state, and so it made sense at the time. 
After implementing this geo-filter, we realized that none of the data was from Colorado. We wanted some data to test with, so we decided to un-do the geo-filter. 
Instead of commenting it out or deleting it, we decided to leave it active and unit test it, in case we wanted to re-implement it later. 

Now that we know iCone is geo-filtering the data for us we can remove it

payneBrandon: Doesn't the `or True` portion here allow all data through regardless if its in Colorado or not? Moot since it will be removed I suppose though.

jacob6838: Right, we just wanted to disable it in a way that would be easy to re-enable later. Not best practice for sure but it made sense at the time

payneBrandon: Ok, I'm up to speed with you now. Thanks!

translator/source_code/icone_translator.py
payneBrandon: Well that makes more sense now. Not the most helpful thing for xmltodict to behave differently with the data like that, but at least I see why the extra case handling here.

README.md
jacob6838: There is a pip install line in the readme, `pip install -r requirements.txt`. If you run that line (from root) and then try the coverage line, does it work?

payneBrandon:
Not sure why this was closed, @abinashkonersman is this still supposed to be active?

README.md
payneBrandon: @jacob6838, I don't see a reference to "coverage" in the requirements.txt. But I did run `pip install -r requirements.txt` and the coverage call doesn't work. I'm running Windows 10 if that matters, maybe there's an extra step to installing coverage on Windows?

payneBrandon: Note I can run `python -m pytest .\test\ -v` successfully, though it does add a `__pycache__` directory so we'll need to update the .gitigore file for the standard Python set.

jacob6838: It should be installed with `pytest-cov`, hmm...

README.md
payneBrandon: Looks good!

README.md
jacob6838: My mistake, coverage is something else. You are right, the command should be `python -m pytest .\test\`. We just updated this in the main readme

translator/GCP_cloud_function/README.md
abinashkonersman: I removed the Zip file and added the uncompressed files and updated the README.md description

payneBrandon:
I noticed running the test seems to modify the `translator\sample files\Output Message\icone_to_wzdx_test.geojson` file. Could we add this to the .gitignore file?

abinashkonersman:
> I noticed running the test seems to modify the `translator\sample files\Output Message\icone_to_wzdx_test.geojson` file. Could we add this to the .gitignore file?

Good catch ! just added it to gitignore.

.gitignore
```diff
@@ -2,4 +2,6 @@
 **/.idea/*
 .idea/*
 **/__pycache__/**
-.DS_Store
\ No newline at end of file
+.DS_Store
+icone_to_wzdx_test.geojson
```
payneBrandon, line 6: Since this is already part of the solution, you'll have to delete this file and commit that before the gitignore takes affect.

abinashkonersman: just deleted the file

translator/source_code/icone_translator.py
abinashkonersman: The code has been removed.

translator/source_code/icone_translator.py
abinashkonersman: The code has been removed.

translator/source_code/icone_translator.py
payneBrandon: Closing for now, but I think we'll want to look at this logic a bit closer in the next sprint. We could easily have a valid sign on the side of the road that would have the same start/end points since its just a single location. I'm not sure this would have a direction or not, but if not we'll have to determine how to best handle it.

.gitignore
payneBrandon: Looks good!

payneBrandon (approved):
Looks good to me!

PR #15, opened 2021-02-02
Dev

abinashkonersman:
updated google cloud functions
 - automatically downloads the data from ftp server
 - cloud scheduler triggers cloud function daily
 - publishes the translated message to a pub/sub topic


translator/GCP_cloud_function/cloud_function/requirements.txt
```diff
@@ -1,6 +1,9 @@
 # Function dependencies, for example:
 # package>=version
+# Function dependencies, for example:
+# package>=version
```
payneBrandon: Duplicate, can we remove?

abinashkonersman: i removed the duplicates.

translator/GCP_cloud_function/cloud_function/main.py
```diff
@@ -1,35 +1,53 @@
...
 import json
 import xmltodict
+import shutil
+import urllib.request as request
+from contextlib import closing
+from google.cloud import pubsub_v1
+
+
+server = 'iconetraffic.com'
+port = '42663'
```
payneBrandon: These values shouldn't be hard coded, but rather come from configuration variables. [Environment Variables](https://cloud.google.com/functions/docs/env-var) may be a good choice

translator/GCP_cloud_function/cloud_function/main.py
```diff
@@ -1,35 +1,53 @@
...
+user = 'cdot'
+password = 'icone_cdot'
```
payneBrandon: Any potentially sensitive information, such as username and password, should not be hard coded in here. Instead these should be coming from a secure location, try [Secrets](https://dev.to/googlecloud/using-secrets-in-google-cloud-functions-5aem).

translator/GCP_cloud_function/cloud_function/main.py
```diff
@@ -1,35 +1,53 @@
...
+
+  icone_data=get_ftp_file(ftpUrl)
+  icone_obj=parse_xml(icone_data)
+  info=icone_translator.initialize_info()
+
+  wzdx_obj=icone_translator.wzdx_creator(icone_obj,info)
+  wzdx_schema=get_wzdx_schema('wzdx_schema.json')
+  validation_result=icone_translator.validate_wzdx(wzdx_obj,wzdx_schema)
+  # if not validation_result:
+  #   print('validation failed! see previous messages for error messages')
```
payneBrandon: Dead code, can we remove?

translator/GCP_cloud_function/cloud_function/main.py
```diff
@@ -1,35 +1,53 @@
...
+  print(json.dumps(wzdx_obj))
+
+  publisher = pubsub_v1.PublisherClient()
+  topic_path = publisher.topic_path('cdot-cv-ode-dev', 'wzdx_messages')
```
payneBrandon: These variables should also come from config settings. We'll be changing them in staging and production environments and will not want to have to recompile or change code to adjust environment specific names.

translator/GCP_cloud_function/cloud_function/main.py
payneBrandon: Actually, maybe this should be uncommented? It looks like we get a "validation_result" and then proceed without checking it. We want to be ensuring only valid data gets through.

translator/GCP_cloud_function/cloud_function/main.py
```diff
@@ -1,35 +1,53 @@
...
+user = 'cdot'
+password = 'icone_cdot'
+filepath = 'incidents.xml'
+
+
+ftpString = 'ftp://{0}:{1}@{2}:{3}/{4}'
+ftpUrl = ftpString.format(user, password, server, port, filepath)
+
+
+def hello_pubsub(event, context):
```
payneBrandon: Let's name this something meaningful so we know what its purpose is.

translator/GCP_cloud_function/cloud_function/main.py
```diff
@@ -1,35 +1,53 @@
...
+
+
+  icone_data=get_ftp_file(ftpUrl)
+  icone_obj=parse_xml(icone_data)
+  info=icone_translator.initialize_info()
```
payneBrandon: This info object seems to be both produced and consumed by the icon_translator file. Can we just leave it in there and not fetch and repush the object?

translator/GCP_cloud_function/README.md
```diff
@@ -1,9 +1,10 @@
 # iCone translator Cloud Function
 
-A system was created in google cloud platform to automatically translate iCone data to WZDx message.This system consits of two buckets and a cloud function.
-When a file is uploaded to the first bucket, the cloud function automatically downloads the file, translates to a WZDx message and uploads it to the second bucket.
+A system was created in google cloud platform to automatically translate iCone data to WZDx message. This system consists of two pubsub topics and a cloud function.
```
payneBrandon, line 3: I'm only seeing the single pubsub topic that we are sending data to at this point. Looking at `main.py` it appears that the current logic is as follows:
1. hello_pubsub triggered (not sure how)
2. call out to FTP site and pull file
3. parse file to xml
4. translate xml to wzdx object
5. call validation function
6. publish wzdx object to pubsub

In general, we shouldn't be publishing to a pubsub topic from a scheduler just to kick off a cloud function. We can set the scheduler to call the function directly.

payneBrandon:
Once you get the environment variables going, let's also add in variables for contact_name and contact_email in the `icone_translator.py`.

translator/GCP_cloud_function/cloud_function/icone_translator.py
```diff
@@ -136,33 +106,30 @@ def get_road_direction(coordinates):
...
-    if street_char == 'N':
-        direction='northbound'
-    elif street_char =='S':
-        direction ='southbound'
-    elif street_char =='W':
-        direction ='westbound'
-    elif street_char =='E':
-        direction ='eastbound'
+    # function to parse direction from street name
+    street_char = street[-1]
```
payneBrandon, line 120: Do we still need the check in here for empty street? What happens when you pass this function an empty string?

translator/GCP_cloud_function/cloud_function/icone_translator.py
```diff
@@ -1,16 +1,12 @@
 import xmltodict
```
payneBrandon: It doesn't look like we're using xmltodict in this file, can we remove?

payneBrandon:
`icone_translator.py` line 313 - 317 deals with start/end mileposts. If we're not including these, can we remove the commented code?

translator/GCP_cloud_function/cloud_function/icone_translator.py
```diff
@@ -372,48 +407,17 @@ def initialize_info():
...
-        return False
-
-    with open(outputfile, 'w') as fwzdx:
-        fwzdx.write(json.dumps(wzdx_obj, indent=2))
-
-    # if not os.path.exists(outputfile) or os.stat(outputfile).st_size == 0:
-    #     print('Output file creation failed . file does not exist or is empty')
+def validate_wzdx(wzdx_obj, wzdx_schema):
+    #wzdx_schema = json.loads(open(location_schema).read())
+    # try:
```
payneBrandon: Should this `try` still be in the file? If not and we come upon an invalid file, it will throw an error and cause an exception.

translator/GCP_cloud_function/cloud_function/main.py
abinashkonersman: We removed the commented code.
This will throw an exception if validation fails, so there is no reason to check the result. We removed the variable assignment to reflect that.

translator/GCP_cloud_function/cloud_function/main.py
abinashkonersman: i renamed it to translate_newest_icone_to_wzdx

translator/GCP_cloud_function/cloud_function/main.py
abinashkonersman: These were transitioned to environment variables.

translator/GCP_cloud_function/cloud_function/main.py
abinashkonersman: We did not have permissions to access secrets ,but we will pursue it in next sprint.

translator/GCP_cloud_function/cloud_function/main.py
abinashkonersman: These were transitioned to environment variables.

translator/GCP_cloud_function/cloud_function/main.py
payneBrandon: What happens when the `validate_wzdx` function throws an exception? Does the whole app crash? Do we get any logging to help troubleshoot what went wrong? We need to be checking the validation and logging any issues so we can figure out what went wrong in the future.

translator/GCP_cloud_function/cloud_function/requirements.txt
payneBrandon: These duplicates are still here, or were added in again with a recent push.

translator/GCP_cloud_function/cloud_function/main.py
```diff
@@ -1,35 +1,69 @@
...
+    return r.read().decode('utf-8-sig')
+
+
+def parse_xml(xml_string):
+  return xmltodict.parse(xml_string)
+
+def get_wzdx_schema(schema_file_name):
+  return json.loads(open(schema_file_name,'r').read())
+
+def get_ftp_credentials():
```
payneBrandon, line 56: Is this function needed? It looks like the call to it is commented out above and you're accessing the properties directly instead.

translator/GCP_cloud_function/cloud_function/main.py
```diff
@@ -1,35 +1,69 @@
...
+import os
+from google.cloud import secretmanager
+
+
+
+server = os.environ['ftp_server_address']
+port = os.environ['ftp_port']
+user=os.environ['icone_ftp_username']
+password=os.environ['icone_ftp_password']
+#user,password=get_ftp_credentials()
```
payneBrandon, line 18: Dead code, can we remove?

translator/GCP_cloud_function/cloud_function/main.py
abinashkonersman: we pre-created it for when we have secrets working and we are going to start using this function when we get permissions to use secrets.

translator/GCP_cloud_function/cloud_function/main.py
abinashkonersman: we are going to start using this function when we get permissions to use secrets.

translator/GCP_cloud_function/cloud_function/main.py
payneBrandon: Let's at least add a comment to the start of the function explaining that it is a function to parse out secrets, and is to be implemented.

translator/GCP_cloud_function/cloud_function/icone_translator.py
abinashkonersman: we removed the commented code and we are erring on the side of only pushing valid data .if validation failed ,we want to throw an exception.

payneBrandon: So what happens when an exception is thrown? It doesn't look like we are handling it, or logging anything.

translator/GCP_cloud_function/cloud_function/icone_translator.py
abinashkonersman: we removed it.

translator/GCP_cloud_function/cloud_function/icone_translator.py
abinashkonersman: we added a case if street is empty. In practice this will never happen because this function will never be called if street is empty.

translator/GCP_cloud_function/README.md
abinashkonersman: The second pub/sub topic is for triggering the function.we didn't know ,we could trigger it directly from cloud scheduler ,we will remove the second pub/sub topic in next sprint.

translator/GCP_cloud_function/cloud_function/main.py
abinashkonersman: This was made a bit awkward because we need to set the info object to specific values for all of our work, but if another person wanted to use the translator then they would want to create their own custom info object. We should not have the initialize_info() function in the icone_translator, but should be doing that on our own. due to time constraints, we will likely only get to this next sprint.

translator/GCP_cloud_function/cloud_function/main.py
jacob6838: The cloud function will crash and stop execution. The ValidationError thrown by the validate_wzdx function has all of the information about what field failed validation and why, making it (usually) easy to diagnose the issue. It looks like we can set up alerts (email or text) for if the function fails, that would be a good thing to do.

payneBrandon: We really should be handling any exceptions that would cause the entire app to crash. The exception should be caught, logged to output, and then allow the app to exit gracefully.

translator/GCP_cloud_function/cloud_function/requirements.txt
abinashkonersman: I just pushed the update, they should be removed now.

translator/GCP_cloud_function/cloud_function/main.py
jacob6838: The way I am thinking about it, I want the app to crash and throw an error, so that we can easily tell when it fails. If we simply log an error message, then it is much less obvious that something went wrong. 
My thinking is that if the WZDx message validation fails, then the WZDx message is invalid and an uncaught error has occurred somewhere, meaning that the function has failed

jacob6838: If we decide that we still want to publish an invalid WZDx message, than we should definitely be handling the error. Otherwise, we need to be sounding alarm bells because something broke, and the easiest way I can see to do that is to crash the function app, leaving traceback info in the cloud function logs.

payneBrandon: If we want to raise alarm bells for an invalid message (which makes sense as this is an odd issue), we should add in some notification logic to either send out an email or fire off some other trigger. The app itself should never be designed to fail though. With current logic allowing the app to crash, is there some notification we have set up to alert on cloud function fail, or would the GCP manager see this on a dashboard view or another location?

translator/GCP_cloud_function/cloud_function/main.py
payneBrandon: You could make the `info` object optional, then check if it's null on the translator side. If it's null, apply your default values produced by the `initialize_info` method.

If another application is going to call the translator, that will allow them to pass in custom values while avoiding the double call on our end here.

translator/GCP_cloud_function/README.md
payneBrandon: @abinash-konersman can you write up a story for this and throw a link to it here so we have it documented.

translator/GCP_cloud_function/cloud_function/main.py
jacob6838: great idea

translator/GCP_cloud_function/cloud_function/icone_translator.py
payneBrandon: Ah, now I see the road_name check above. We can make that code a bit clearer on line 279 - 290 by reusing the "road_name" variable and changing as follows:
```
# road_name
    road_name = incident['location'].get('street', '')
    if not road_name:
        return None
    properties['road_name'] = road_name

    # direction
    direction = parse_direction_from_street_name(road_name)
    if not direction:
        direction = get_road_direction(geometry['coordinates'])
        if not direction:
            return None
```

translator/GCP_cloud_function/cloud_function/main.py
jacob6838: @abinashkonersman Can you write up a story for improving the error reporting (and not crashing the app)? this link seems relevant: https://cloud.google.com/functions/docs/monitoring/error-reporting

translator/GCP_cloud_function/cloud_function/main.py
abinashkonersman: I commented it in code.Thanks

translator/GCP_cloud_function/README.md
payneBrandon: After working with Dhivahari a bit on the pubsub/scheduler stuff you are correct here. It appears the best way to get the scheduler to talk to the cloud function is sending a message through pubsub (which just seems very strange, but it's what the documentation says to do!). No need to write up the story mentioned above. Let's do get the README in line with what the code is actually doing though.

translator/GCP_cloud_function/cloud_function/icone_translator.py
payneBrandon: Looks like https://cdot-its.atlassian.net/browse/CV-64?atlOrigin=eyJpIjoiMDA3OTAwZDFlYzNkNDJiZmFkMTQ5ZWJkYjBkMmIyNDQiLCJwIjoiaiJ9 will handle this. Closing

translator/GCP_cloud_function/cloud_function/main.py
payneBrandon: https://cdot-its.atlassian.net/browse/CV-64?atlOrigin=eyJpIjoiMDA3OTAwZDFlYzNkNDJiZmFkMTQ5ZWJkYjBkMmIyNDQiLCJwIjoiaiJ9 Seems to have been written to handle this. Closing.

translator/GCP_cloud_function/cloud_function/main.py
abinashkonersman: The info object is now optional

translator/GCP_cloud_function/cloud_function/icone_translator.py
abinashkonersman: I just made  your requested change.

payneBrandon (approved):
Looks good.

PR #19, opened 2021-02-16
Dev

abinashkonersman:
-Added Unit test to GCP code
-Updated README.md in gcp directory


abinashkonersman:
please review my code

sonar-project.properties
```diff
@@ -11,7 +11,7 @@ sonar.links.homepage=https://github.com/CDOT-CV/Work_Zone
 sonar.links.scm=https://github.com/CDOT-CV/Work_Zone
 sonar.links.issue=https://github.com/CDOT-CV/Work_Zone/issues
 sonar.python.coverage.reportPaths=*coverage*.xml
-sonar.exclusions=*coverage*.xml,test/*,translator/GCP_cloud_function/**
+sonar.exclusions=*coverage*.xml,test/*,translator/GCP_cloud_function/cloud_function/icone_translator.py
```
payneBrandon: Is the `icone_translator.py` under "cloud_function" directory the same as the one under "source_code" directory? If so, let's just have the single code file and import it to the cloud function. Then we can remove the one under the "cloud_function" directory and not have to do this odd exclusion.

payneBrandon:
Unit tests are failing

payneBrandon:
This PR is out of date with the main branch. You'll need to pull in the changes from `main` locally and push those up to dev.

sonar-project.properties
abinashkonersman: Removed icone_translator.py from "GCP_cloud_function" directory and mader recomended changes and pushed it to dev.

abinashkonersman:
> Unit tests are failing

@payneBrandon can you please check again i just recently pushed the updated code and let me know if unit tests are working or not .

payneBrandon:
> > Unit tests are failing
> 
> @payneBrandon can you please check again i just recently pushed the updated code and let me know if unit tests are working or not .

Still seeing errors when attempting to run `python -m pytest 'test/' -v`. Here's the output that I'm seeing: 
image

sonar-project.properties
payneBrandon: Let's remove the icone_translator.py sonar exclusion as well. The file no longer exists, so no need to exclude it.

translator/GCP_cloud_function/cloud_function/main.py
```diff
@@ -52,18 +49,17 @@ def parse_xml(xml_string):
 def get_wzdx_schema(schema_file_name):
   return json.loads(open(schema_file_name,'r').read())
 
-#this function retrieves username and password info from secrets .This will be implemented after correct secrets are implemented.
 def get_ftp_credentials():
   secret_client = secretmanager.SecretManagerServiceClient()
-  username_secret_name = os.environ['icone_ftp_username_secret_name']
-  password_secret_name = os.environ['icone_ftp_password_secret_name']
+  username_secret_name = os.environ['icone_ftp_username_secret_name']#"icone_ftp_username"
```
payneBrandon: Didn't we remove the comments from here before?

test/test_translator.py
```diff
@@ -232,26 +229,10 @@ def test_get_event_status():
...
-    # info['metadata']['timestamp_metadata_update'] = "timestamp_metadata_update"
-    # info['metadata']['contact_name'] = "contact_name"
-    # info['metadata']['contact_email'] = "contact_email"
-    # info['metadata']['issuing_organization'] = "issuing_organization"
-    test_wzdx=icone_translator.wzdx_creator(icone_obj,icone_translator.initialize_info())
-
-    validate(instance=test_wzdx,schema=wzdx_schema)
+    test_wzdx = icone_translator.wzdx_creator(icone_obj, icone_translator.initialize_info())
+    validate(instance=test_wzdx, schema=wzdx_schema)
+    assert True
```
payneBrandon: We shouldn't be asserting True, that doesn't prove our test accomplished anything. Rather, we should be validating that the returned object has expected properties. You opened a known file and parsed the xml. That means we know what the properties should look like on the returned object. Verify those properties instead, or we could have a function in here that returns a pre-built wzdx object that we could compare the returned object against.

test/test_translator.py
```diff
@@ -232,26 +229,10 @@ def test_get_event_status():
...
-    # info['metadata']['location_verify_method'] = "location_verify_method"
-    # info['metadata']['datafeed_frequency_update'] = 86400
-    # info['metadata']['timestamp_metadata_update'] = "timestamp_metadata_update"
-    # info['metadata']['contact_name'] = "contact_name"
-    # info['metadata']['contact_email'] = "contact_email"
-    # info['metadata']['issuing_organization'] = "issuing_organization"
-    test_wzdx=icone_translator.wzdx_creator(icone_obj,icone_translator.initialize_info())
-
-    validate(instance=test_wzdx,schema=wzdx_schema)
+    test_wzdx = icone_translator.wzdx_creator(icone_obj, icone_translator.initialize_info())
```
payneBrandon: This is the only test against the `wzdx_creator()` function, which means the only case we are testing is a valid icon_obj, and valid info object. We want to test all the edge cases - what happens if we pass in a null icone_obj, or no info object for instance?

test/test_translator.py
```diff
@@ -4,9 +4,6 @@
 import xmltodict
 import json
 from jsonschema import validate
-from datetime import datetime
-import subprocess
-
```
payneBrandon, line 11: We're still missing edge case testing in this file overall. We need to add in cases to test where no data is passed to the function, or bad data is.

Also, unit tests should be named so we know what they are testing and not just the function being tested. For example the polyline test below we could name `test_parse_polyline_valid_points` then if we test with no data, `test_parse_polyline_no_data`.

test/test_main.py
```diff
@@ -0,0 +1,81 @@
...
+@patch('urllib.request')
+def test_get_ftp_file(request):
+    test_url='fake url'
+    test_output=main.get_ftp_file(test_url)
+    request.urlopen.assert_called_with(test_url)
+    assert True
+
+
+@patch('google.cloud.pubsub_v1.PublisherClient')
+@patch.object(main, 'get_ftp_url')
```
payneBrandon, line 25: Since you're setting the `main.xxx = MagicMock()` below, we don't need to `@patch.object` these in, or include them as parameters.

test/test_main.py
```diff
@@ -0,0 +1,81 @@
...
+@patch.object(main, 'get_ftp_file')
+@patch.object(main, 'parse_xml')
+@patch.object(main, 'get_wzdx_schema')
+def test_translate_newest_icone_to_wzdx(get_wzdx_schema, parse_xml, get_ftp_file, get_ftp_url, pubsub):
```
payneBrandon: This is a good first test with valid data. We need to add additional tests for cases with invalid data. For instance, when `get_wzdx_schema` returns empty or `validate_wzdx` returns false.

test/test_main.py
```diff
@@ -0,0 +1,81 @@
...
+    icone_translator.validate_wzdx= MagicMock(return_value=True)
+    os.environ['project_id']='project_id'
+    os.environ['wzdx_topic_id']='wzdx_topic_id'
+    main.translate_newest_icone_to_wzdx(None,None)
+    publisher=pubsub().publish
+    publisher.assert_called_with(pubsub().topic_path('project_id', 'wzdx_topic_id'),str.encode(json.dumps('WZDx', indent=2)),origin='auto_icone_translator_ftp cloud function')
+    assert True
+
+
+@patch.object(main, 'get_ftp_credentials')
```
payneBrandon, line 110: Again, we're using the direct setting of MagicMock below, so don't need this patch.

test/test_main.py
```diff
@@ -0,0 +1,81 @@
...
+def test_get_ftp_url(ftp_credentials):
```
payneBrandon, line 111: Good first test with this method, but we still need the edge cases. What happens if an environment variable isn't set?

translator/GCP_cloud_function/cloud_function/main.py
```diff
@@ -42,8 +39,8 @@ def translate_newest_icone_to_wzdx(event, context):
   return
 
 def get_ftp_file(url) :
-  with closing(request.urlopen(url)) as r:
-    return r.read().decode('utf-8-sig')
+  return urllib.request.urlopen(url).read().decode('utf-8-sig')
+
 
 
 def parse_xml(xml_string):
```
payneBrandon, line 52: This `parse_xml` method is really just wrappinga single functionality - a call to the xmltodict library. Since we aren't doing anything else in here we aren't gaining anything by using this abstraction. Let's remove in favor of directly calling the `xmltodict.parse()` method.

translator/GCP_cloud_function/cloud_function/main.py
```diff
@@ -52,18 +49,17 @@ def parse_xml(xml_string):
 def get_wzdx_schema(schema_file_name):
```
payneBrandon, line 55: Similar to `parse_xml`, this function isn't gaining us much. It is doing a little more with opening the file, then parsing it in json so it may be worth leaving in. If we do leave it in, we need to write unit tests for it to show what happens with a valid file, and what happens if the file doesn't exist.

test/test_main.py
```diff
@@ -0,0 +1,81 @@
...
+          <polyline>34.8380671,-114.1450650,34.8380671,-114.1450650</polyline>
+        </incident>"""
+
+    test_valid_output= {"incident": OrderedDict({'@id': 'U13631595_202012160845', 'updatetime': '2020-12-16T17:18:00Z', 'type': ['CONSTRUCTION', 'Hazard'], 'polyline': '34.8380671,-114.1450650,34.8380671,-114.1450650' })}
+    actual_output= main.parse_xml(test_parse_xml_string)
+    assert actual_output == test_valid_output
+
+
+@patch('google.cloud.secretmanager.SecretManagerServiceClient')
+def test_get_ftp_credentials(secret):
```
payneBrandon, line 144: Again a good start where all the data is perfect. Let's add those edge cases in though - what happens when the environment variable is absent? or if `access_secret_version` returns invalid data (ie secret doesn't exist)?

test/test_main.py
```diff
@@ -0,0 +1,81 @@
...
+    main.parse_xml=MagicMock(return_value='')
+    main.get_wzdx_schema=MagicMock(return_value='')
+    icone_translator.wzdx_creator= MagicMock(return_value='WZDx')
+    icone_translator.validate_wzdx= MagicMock(return_value=True)
+    os.environ['project_id']='project_id'
+    os.environ['wzdx_topic_id']='wzdx_topic_id'
+    main.translate_newest_icone_to_wzdx(None,None)
+    publisher=pubsub().publish
+    publisher.assert_called_with(pubsub().topic_path('project_id', 'wzdx_topic_id'),str.encode(json.dumps('WZDx', indent=2)),origin='auto_icone_translator_ftp cloud function')
+    assert True
```
payneBrandon: We're not gaining anything by assertng a boolean, let's remove.

test/test_main.py
```diff
@@ -0,0 +1,81 @@
...
+    os.environ['icone_ftp_username_secret_name']='secret_username'
+    os.environ['icone_ftp_password_secret_name']='secret_password'
+    os.environ['project_id'] = 'project_id'
+    main.get_ftp_credentials()
+    valid_secret_user_request={"name": "projects/project_id/secrets/secret_username/versions/latest"}
+    valid_secret_pass_request = {"name": "projects/project_id/secrets/secret_password/versions/latest"}
+    requests=[call(valid_secret_user_request),call().payload.data.decode("UTF-8"), call(valid_secret_pass_request), call().payload.data.decode("UTF-8")]
+    secret_client=secret().access_secret_version
+    secret_client.assert_has_calls(requests)
+    assert True
```
payneBrandon: We're not gaining anything by assertng a boolean, let's remove.

abinashkonersman: I removed it and pushed the updated code.

test/test_main.py
```diff
@@ -0,0 +1,81 @@
...
+import os
+import json
+
+
+@patch('urllib.request')
+def test_get_ftp_file(request):
+    test_url='fake url'
+    test_output=main.get_ftp_file(test_url)
+    request.urlopen.assert_called_with(test_url)
+    assert True
```
payneBrandon: We're not gaining anything by asserting a boolean, let's remove.

test/test_main.py
```diff
@@ -0,0 +1,81 @@
...
+def test_get_ftp_url(ftp_credentials):
+    credentials='username', 'password'
+    main.get_ftp_credentials=MagicMock(return_value=credentials)
+    os.environ['ftp_server_address']='www.icone.com'
+    os.environ['ftp_port']='4425'
+    os.environ['icone_ftp_username']='username'
+    os.environ['icone_ftp_password']='password'
+    os.environ['ftp_icone_file_path']='test_filepath'
+    test_ftp_url='ftp://username:password@www.icone.com:4425/test_filepath'
+    # print(ftp_credentials())
```
payneBrandon: Can we remove?

abinashkonersman: i removed it pushed the code

test/test_main.py
abinashkonersman: i removed it pushed the updaated code.

test/test_main.py
abinashkonersman: i removed it pushed the updated code

test/test_main.py
payneBrandon: Seems to still be here in the [dev branch](https://github.com/CDOT-CV/Work_Zone/blob/db9ae42bc5816f3322e8aec38f16acfd5dc15ed1/test/test_main.py#L18)

test/test_main.py
```diff
@@ -0,0 +1,80 @@
...
+def test_get_ftp_url(ftp_credentials):
+    credentials='username', 'password'
+    main.get_ftp_credentials=MagicMock(return_value=credentials)
+    os.environ['ftp_server_address']='www.icone.com'
+    os.environ['ftp_port']='4425'
+    os.environ['icone_ftp_username']='username'
+    os.environ['icone_ftp_password']='password'
+    os.environ['ftp_icone_file_path']='test_filepath'
+    test_ftp_url='ftp://username:password@www.icone.com:4425/test_filepath'
+    print(main.get_ftp_credentials())
```
payneBrandon: Do we need to print this out during a test? What are we gaining?

abinashkonersman: sorry i was testing the what kind of data i am getting from that function..so mistakenly i forgot to remove the print statement.I removed and pushed the updated code

test/test_main.py
abinashkonersman: sorry for that and i removed I just pushed the updated code to dev again.

test/test_main.py
jacob6838: This is case where it is more complicated than it seems. If we remove the patch.object line, then the main.xxx = MagicMock() line will overwrite the actual function from main, in all contexts. If we keep the patch.object line, it only mocks the function within that unit test.

translator/GCP_cloud_function/cloud_function/main.py
jacob6838: The intent of creating a separate function was to make refactoring/updating of the code easier. If in the future we decide to use a different xml parsing library or we want to add more parameters to the xmltodict.parse function call, then we only need to touch the parse_xml function, and we have unit tests that will ensure that the xml data is being parsed as expected

translator/GCP_cloud_function/cloud_function/main.py
jacob6838: This was added to make unit testing easier

test/test_translator.py
jacob6838: The validate function (from the jsonschema library) throws an exception if the validation fails. So if the validation succeeds, it will assert True and pass, if the validation fails then 
There are some random values created in the wzdx_creator, such as a UUID and two uses of datetime.now(). I do not know of a way to compare an object containing random values to a pre-build object that doesn't feel hacky, that is why we chose to use the validate function instead of equating the output.

test/test_translator.py
abinashkonersman: Added some more edge cases and pushed the updated code to dev

test/test_main.py
abinashkonersman: Added some more edge cases and created a new test function and pushed the updated code to dev.
Thanks

test/test_main.py
abinashkonersman: the main.xxx = MagicMock() line will overwrite the actual function from main, in all contexts. If we keep the patch.object line, it only mocks the function within that unit test.

test/test_main.py
abinashkonersman: added some more edge cases to test_get_ftp_url test function according to your instruction and pushed the updated code to dev.

test/test_translator.py
payneBrandon: Validating the schema here is good, in that it is showing us the return value is a valid WZDx 3.0 schema. We don't need to `assert True` with that, but if we want to assert anything maybe we can assert no exception thrown? 

Either in this method or another unit test, this function returns an object that we expect to have certain properties and we should test for that. The test currently is just saying "any wzdx 3.0 message will do, regardless of the data I've given you". For instance, we are passing in the `icone_translator.initialize_info` as the "info" parameter. We could create a unit test version of this object (static so we know values), and verify that the output object contains expected values. IE our `test_wzdx` object has a feed_info_id set from that info, as well as some metadata, etc.

Also, we are currently reading in the `incidents_exended.xml` file and casting it to a dictionary with the `xmltodict.parse` method. Rather, we could have a known dictionary with data to test return values against.

test/test_translator.py
```diff
@@ -1,17 +1,13 @@
...
+import jsonschema
 
 
 
 
-#Unit testing code
+#Unit testing code for icone_translator.py
 def test_parse_incident() :
```
payneBrandon: Can we break these unit tests out to test a single unit each? This one seems to be testing a success case, and a case where we don't find a direction. Both are valid tests, and should be tested individually with appropriate names to the test. For instance the success case could be called `test_parse_incident_success` and the direction tests `test_parse_incident_no_direction` or similar.

There's several additional edge cases to test with this `parse_incident` method as well, for instance what happens when we pass `None` to the method? Or invalid data?

test/test_translator.py
```diff
@@ -138,6 +117,36 @@ def test_parse_polyline() :
         ]
     assert  test_coordinates == valid_coordinates
 
+def test_parse_polyline_invalid_data() :
```
payneBrandon: Each of these asserts seems to be testing a valid unit. We should break those out into appropriatly named unit tests of their own.

test/test_translator.py
```diff
@@ -138,6 +117,36 @@ def test_parse_polyline() :
...
+
+
+
+
+
+
+    
+
 
 def test_get_road_direction():
```
payneBrandon: Same as above, these should to be broken out to test a single unit of logic with appropriate names to the methods so we know what is being tested.

test/test_translator.py
```diff
@@ -138,6 +117,36 @@ def test_parse_polyline() :
...
     test_coordinates = [
```
payneBrandon, line 280: We'll want additional tests to show what happens when we pass this function an empty set, None, and invalid data.

translator/GCP_cloud_function/cloud_function/main.py
abinashkonersman: added edge case for  get_wzdx_schema and for invalid data i named the function test_get_wzdx_schema_invalid_data() and pushed the updated code to dev

test/test_translator.py
abinashkonersman: I used a regular expression to compare the generated wzdx message to the hard coded one.The reason i used the regular expression was to match the randomly generated  fields such as id and added some unity test edge cases for the function and pushed the updated code to dev branch.

test/test_translator.py
abinashkonersman: Added some unity test edge cases for the function and pushed the updated code to dev branch.

test/test_translator.py
abinashkonersman: added additional edge cases with valid test function name and pushed into dev.

test/test_translator.py
abinashkonersman: added additional edge cases with valid test function name and pushed into dev.

test/test_translator.py
abinashkonersman: added valid test function name and pushed into dev.

test/test_translator.py
abinashkonersman: added additional edge cases with valid test function name and pushed into dev.

test/test_translator.py
```diff
@@ -84,27 +81,29 @@ def test_parse_incident() :
     ]
   }
 }
+    assert test_feature == valid_feature
 
+def test_parse_incident_success_no_data():
```
payneBrandon: Since we're not testing a "successful" case here can we name it something more appropriate? Maybe `test_parse_incident_no_data` where we just drop the "success" text as it's missleading.

test/test_translator.py
```diff
@@ -84,27 +81,29 @@ def test_parse_incident() :
...
+  test_feature = icone_translator.parse_incident(None)
+  valid_feature=None
+  assert test_feature == valid_feature
 
-    assert test_feature == valid_feature
+def test_parse_incident_success_invalid_data():
```
payneBrandon: Instead of throwing a random error here we should be testing that the callback we included is called:
```
def test_parse_incident_success_invalid_data():
    test_var = 'a,b,c,d'
    callback = MagicMock()
    test_feature = icone_translator.parse_incident(
        test_var, callback_function=callback)
    assert callback.called
```

Also, we aren't testing a success case here, so let's remove that from the name.

payneBrandon: We'll also want to verify the returned `test_feature` here is our expected value (None in this case).

test/test_translator.py
```diff
@@ -1,18 +1,16 @@
...
+import re
 
 
 
 
+#Unit testing code for icone_translator.py
 
-#Unit testing code
-def test_parse_incident() :
+def test_parse_incident_success() :
```
payneBrandon: Here we test where the incident has direction from the street, which is a valid test case. We should name this method to recognize this case and have another test that hits the case for when we calculate direction based on coordinates.

translator/source_code/icone_translator.py
```diff
@@ -353,6 +364,34 @@ def parse_incident(incident):
...
+        incident['starttime']
+        incident['description']
+        incident['creationtime']
+        incident['updatetime']
+        road_name = incident['location']['street']
+        direction = parse_direction_from_street_name(road_name)
+        if not direction:
+            direction=get_road_direction(coords)
+            if not direction:
+                raise RuntimeError('unable to parse direction from street name or polyline')
```
payneBrandon: Rather than raise an exception here (which we just catch right below), we should do similar to lines 371-372 - log a warning, and return False. Raising an exception is typically very slow, and often unnecessary. A faster approach is just to return False and avoid the exception.

translator/source_code/icone_translator.py
```diff
@@ -353,6 +364,34 @@ def parse_incident(incident):
...
 
+#function to validate the incident
+def validate_incident(incident):
+    
+    if not incident or (type(incident) != dict and type(incident) != OrderedDict):
+        logging.warning('incident is empty or has invalid type')
+        return False
+    try:
+        coords = parse_polyline(incident['location']['polyline'])
+        incident['starttime']
```
payneBrandon: Rather than relying on this dictionary to thrown an error if the key isn't found (which again is slow), we should check for the existence ourselves:
```
starttime = incident.get('starttime')
```
Then we can check if starttime is None or a value.

test/test_translator.py
```diff
@@ -138,8 +136,73 @@ def test_parse_polyline() :
         ]
     assert  test_coordinates == valid_coordinates
 
+def test_parse_polyline_empty_string() :
```
payneBrandon: Since we're not testing an empty string value here (''), we should name this test something like `test_parse_polyline_null_parameter` or simliar.

test/test_translator.py
```diff
@@ -138,8 +136,73 @@ def test_parse_polyline() :
...
+
+    test_polyline= None
+    test_coordinates=icone_translator.parse_polyline(test_polyline)
+    valid_coordinates= []
```
payneBrandon: This naming is a bit misleading. These aren't valid coordinates exactly, they are just the expected coordinates. Can we rename this to 'expected_coordinates'?

test/test_translator.py
```diff
@@ -138,8 +136,73 @@ def test_parse_polyline() :
...
+    assert  test_coordinates == valid_coordinates
+
+def test_parse_polyline_invalid_data() :
+    test_polyline= 'invalid' 
+    test_coordinates=icone_translator.parse_polyline(test_polyline)
+    valid_coordinates= []
+    assert  test_coordinates == valid_coordinates
 
-def test_get_road_direction():
+def test_parse_polyline_invalid_coordinates() :
```
payneBrandon: Rather than `assert True` or `assert False` here we should be using built-in testing tools to check this exception for us:
```
def test_parse_polyline_invalid_coordinates():
    test_polyline = 'a,b,c,d'
    with pytest.raises(RuntimeError) as runtimeErr:
        icone_translator.parse_polyline(test_polyline)
    assert "Failed to parse polyline data." in str(runtimeErr.value)
```
Here we import pytest, then we can use it to verify the exception is raised. Then we can even go so far as to assert the exception message is what we expected it to be.

Alternatively, we could wrap these tests in a class and pass that class "unittest.TestCase" then the test method can take in "self" and call "self.assertRaises":
```
class TestTranslator(unittest.TestCase):

    def test_parse_polyline_invalid_coordinates(self):
        test_polyline = 'a,b,c,d'
        with self.assertRaises(RuntimeError) as runtimeErr:
            icone_translator.parse_polyline(test_polyline)
```

This method has a big benefit of the ".assertRaises" which is asserting the exception did in fact fire off.

test/test_translator.py
```diff
@@ -138,8 +136,73 @@ def test_parse_polyline() :
...
+    'description': 'Road constructions are going on',
+    'creationtime': '2020-12-13T14:18:00Z', 
+    'location': {
+      'polyline': '34.8380671,-114.1450650,34.8380671,-114.1450650',
+      'street': 'I-70 N'
+    }
+  }
+  assert icone_translator.validate_incident(test_valid_output) == False
+
+def test_validate_incident_no_data():
```
payneBrandon, line 273: Can we add another case for passing a value that is not None but is also not `dict` or `OrderedDict`?

test/test_translator.py
```diff
@@ -138,8 +136,73 @@ def test_parse_polyline() :
...
+    'description': 'Road constructions are going on', 
+    'creationtime': '2020-12-13T14:18:00Z', 
+    'location': {
+      'polyline': '34.8380671,-114.1450650,34.8380671,-114.1450650',
+      'street': 'I-70 N'
+    }
+  }
+  assert icone_translator.validate_incident(test_valid_output) == True
+
+def test_validate_incident_missing_required_field():
```
payneBrandon: Which field are we testing here that is missing? This should be included in the name so we don't have to go looking.

translator/source_code/icone_translator.py
```diff
@@ -77,25 +87,36 @@ def wzdx_creator(messages, info):
...
+            raise RuntimeError('Failed to parse polyline data.') from e
     return coordinates
 
 
 # function to get road direction by using geometry coordinates
 def get_road_direction(coordinates):
-    long_dif = coordinates[-1][0] - coordinates[0][0]
-    lat_dif = coordinates[-1][1] - coordinates[0][1]
+    if not coordinates:
+        return []
```
payneBrandon: The rest of this method is returning a string value, yet here we return an empty array. We should be returning None here.

test/test_translator.py
```diff
@@ -155,6 +218,23 @@ def test_get_road_direction():
 
     assert test_direction==valid_direction
 
+def test_get_road_direction_invalid_direction():
```
payneBrandon: We're running three separate tets in this method. Let's break those out into appropriate named methods and test each separately. For instance:
```
def test_get_road_direction_empty_coordinates():
    test_coordinates = []
    test_direction = icone_translator.get_road_direction(test_coordinates)
    expected_direction = None
    assert expected_direction == test_direction
```

test/test_translator.py
```diff
@@ -155,6 +218,23 @@ def test_get_road_direction():
...
+    test_direction=icone_translator.get_road_direction(test_coordinates)
+    valid_direction= []
+    assert test_direction==valid_direction
+
+    test_coordinates = None
+    test_direction=icone_translator.get_road_direction(test_coordinates)
+    valid_direction= []
+    assert test_direction==valid_direction
+
+def test_get_road_direction_valid_direction():
```
payneBrandon: Can we split these unit tests out to test a single unit of logic? IE we should have a test for northbound, a separate test for southbound, eastbound, and westbound.

payneBrandon:
In `test_translator.py` we have a `test_get_vehicle_impact()` method. Currently this is only testing the case that we have a lane closed. We'll want to add in a test for when no lanes are closed as well.

payneBrandon:
Can we split `test_get_event_status()` out to it's three separate units of test? This is a great show of testing the full functionality in that method, but we'll want them to be separate tests.

test/test_translator.py
```diff
@@ -228,31 +308,39 @@ def test_get_event_status():
 
     assert test_event_status == valid_event_status
 
+    
+
 
 def test_wzdx_creator() :
-    wzdx_schema = json.loads(open('translator/sample files/validation_schema/wzdx_v3.0_feed.json').read())
-    icone_data =open('translator/sample files/Icone Data/incidents_extended.xml').read()
```
payneBrandon, line 418: This is a good start for this method, but we still have quite a few cases to cover here. For instance:
 - info object not included (verify it's created)
 - datafeed_frequency_update there vs not
 - no incidents found
 - etc.

test/test_translator.py
```diff
@@ -228,31 +308,39 @@ def test_get_event_status():
...
 
+  assert re.match(wzdx_re,json.dumps(test_wzdx)) != None
+
+def test_wzdx_creator_empty_icone_object() :
+
+  icone_obj = None
+  test_wzdx = icone_translator.wzdx_creator(icone_obj)
+  assert test_wzdx == None
 
 def test_parse_arguments():
```
payneBrandon, line 558: Good success test case, but what about when we have no arguments, or invalid arguments, or the parsing of them fails?

payneBrandon:
There are still a few methods that need additional unit tests written to show full functionality, but getting closer!

test/test_main.py
```diff
@@ -0,0 +1,145 @@
...
+    del os.environ['icone_ftp_username']
+    del os.environ['icone_ftp_password']
+    del os.environ['ftp_icone_file_path']
+    assert actual==test_ftp_url
+
+@patch.object(main, 'get_ftp_credentials')
+def test_get_ftp_url_invalid_data(ftp_credentials):
+    credentials='username', 'password'
+    main.get_ftp_credentials=MagicMock(return_value=credentials)
+    try:
```
payneBrandon: As in `test_translator.py`, we should be using the unit testing tools to verify this exception is thrown rather than surrounding in try/catch

test/test_main.py
```diff
@@ -0,0 +1,145 @@
...
+#the intent of this magic mock fuction is that we give a valid input ,that publishes data
+    main.get_ftp_url=MagicMock(return_value='')
+    main.get_ftp_file=MagicMock(return_value='')
+    main.parse_xml=MagicMock(return_value='')
+    main.get_wzdx_schema=MagicMock(return_value='')
+    icone_translator.wzdx_creator= MagicMock(return_value='WZDx')
+    icone_translator.validate_wzdx= MagicMock(return_value=False)
+    os.environ['project_id']='project_id'
+    os.environ['wzdx_topic_id']='wzdx_topic_id'
+    output=main.translate_newest_icone_to_wzdx(None,None)
```
payneBrandon: Can we verify this output as well?

test/test_main.py
payneBrandon: Can we add a test to handle the exception case as well?

test/test_main.py
```diff
@@ -0,0 +1,145 @@
...
+    test_ftp_url='ftp://username:password@www.icone.com:4425/test_filepath'
+    actual=main.get_ftp_url()
+    del os.environ['ftp_server_address']
+    del os.environ['ftp_port']
+    del os.environ['icone_ftp_username']
+    del os.environ['icone_ftp_password']
+    del os.environ['ftp_icone_file_path']
+    assert actual==test_ftp_url
+
+@patch.object(main, 'get_ftp_credentials')
```
payneBrandon, line 118: Since we are testing missing and not invalid data with this unit test, can we rename to `test_get_ftp_url_missing_environment_variable` or similar?

test/test_main.py
```diff
@@ -0,0 +1,145 @@
...
+    valid_secret_user_request={"name": "projects/project_id/secrets/secret_username/versions/latest"}
+    valid_secret_pass_request = {"name": "projects/project_id/secrets/secret_password/versions/latest"}
+    requests=[call(valid_secret_user_request),call().payload.data.decode("UTF-8"), call(valid_secret_pass_request), call().payload.data.decode("UTF-8")]
+    secret_client=secret().access_secret_version
+    secret_client.assert_has_calls(requests)
+
+
+
+def test_get_wzdx_schema():
+    main.get_wzdx_schema('translator/sample files/validation_schema/wzdx_v3.0_feed.json')
```
payneBrandon: `get_wzdx_schema` returns the schema object. We should  be asserting on that object and not just "True".

test/test_main.py
```diff
@@ -0,0 +1,145 @@
...
+    #wzdx schema will throw an exception if schema is invalid
+    assert True
+
+def test_get_wzdx_schema_invalid_data():
+    try:
```
payneBrandon: Similar to above, let's split this function into it's respective testing units and use the unit testing tools to assert the error is thrown instead of try/catch.

translator/GCP_cloud_function/cloud_function/main.py
```diff
@@ -42,28 +41,59 @@ def translate_newest_icone_to_wzdx(event, context):
...
+    raise RuntimeError('invalid schema: file does not exist')
+  if not schema_string:
+    raise RuntimeError('invalid schema: empty schema')
+  try:
+    schema_obj= json.loads(schema_string)
+    jsonschema.validate('', schema_obj)  
+  except json.decoder.JSONDecodeError as e :
+    raise RuntimeError('invalid schema: not valid json') from e
+  except jsonschema.SchemaError as e :
+    raise RecursionError('invalid schema: schema failed validation') from e
```
payneBrandon: Was this meant to be a RuntimeError? RecursionError is "raised when the interpreter detects that the maximum recursion depth is exceeded". This isn't using recursion so we shoudln't be throwing this error.

test/test_main.py
payneBrandon: We're also checking for four distinct errors in the code, we should have a unit test for each of those checks.

test/test_main.py
```diff
@@ -0,0 +1,145 @@
...
+    main.get_ftp_credentials()
+    valid_secret_user_request={"name": "projects/project_id/secrets/secret_username/versions/latest"}
+    valid_secret_pass_request = {"name": "projects/project_id/secrets/secret_password/versions/latest"}
+    requests=[call(valid_secret_user_request),call().payload.data.decode("UTF-8"), call(valid_secret_pass_request), call().payload.data.decode("UTF-8")]
+    secret_client=secret().access_secret_version
+    secret_client.assert_has_calls(requests)
+
+
+@patch('google.cloud.secretmanager.SecretManagerServiceClient')
+def test_get_ftp_credentials(secret):
```
payneBrandon: Is this just a duplicate of the test method above?

translator/GCP_cloud_function/cloud_function/main.py
```diff
@@ -42,28 +41,59 @@ def translate_newest_icone_to_wzdx(event, context):
...
-  response = client.access_secret_version(request)
+  response = secret_client.access_secret_version(request)
   password = response.payload.data.decode("UTF-8")
 
-  return username,password
\ No newline at end of file
+  return username,password
+
+
+def unsupported_messages_callback(message):
```
payneBrandon, line 101: Can we add some unit tests to show what happens with this method?

test/test_translator.py
abinashkonersman: I changed the name to test_parse_incident_no_data and pushed it to dev

test/test_translator.py
abinashkonersman: I changed the name to test_parse_polyline_null_parameter  and pushed it to dev

test/test_translator.py
abinashkonersman: i renamed it to 
    expected_coordinates= []
    assert  test_coordinates == expected_coordinates

test/test_translator.py
abinashkonersman: i added another edge case :
def test_validate_incident_invalid__output():
  test_valid_output = 'invalid output'
  assert icone_translator.validate_incident(test_valid_output) == False

test/test_translator.py
abinashkonersman: i renamed it to def test_validate_incident_missing_required_field_description():

since in that test case description was missing

test/test_translator.py
```diff
@@ -138,8 +135,86 @@ def test_parse_polyline() :
...
+    'description': 'Road constructions are going on',
+    'creationtime': '2020-12-13T14:18:00Z', 
+    'location': {
+      'polyline': '34.8380671,-114.1450650,34.8380671,-114.1450650',
+      'street': 'I-70 N'
+    }
+  }
+  assert icone_translator.validate_incident(test_valid_output) == False
+
+def test_validate_incident_invalid_test_valid_output():
```
payneBrandon: Let's simplify this name a bit as it's confusing. Maybe `test_validate_incident_invalid`

test/test_translator.py
```diff
@@ -84,27 +84,25 @@ def test_parse_incident() :
...
 
-    icone_obj = xmltodict.parse(test_var)
-    test_feature = icone_translator.parse_incident(icone_obj['incident'])
+def test_parse_incident_invalid_data():
+  test_var = 'a,b,c,d'
+  callback = MagicMock()
+  test_feature = icone_translator.parse_incident(test_var, callback_function=callback)
+  assert callback.called
+
+def invalid_incident_callback(incident):
```
payneBrandon: Now that this is no longer used, can we remove?

test/test_translator.py
```diff
@@ -155,6 +230,23 @@ def test_get_road_direction():
 
     assert test_direction==valid_direction
 
+def test_get_road_direction_empty_coordinates():
```
payneBrandon: We have several distinct tests within this method. The first is testing invalid data, 2nd an empty coordinate set, and 3rd a null coordinate set. These should each be separate tests to show the functionality.

test/test_translator.py
payneBrandon: Nice, let's apply this same thinking to the rest of the unit tests as well.

test/test_translator.py
abinashkonersman: i updated the code and i think i like this concept.

translator/source_code/icone_translator.py
abinashkonersman: i changed the code and it returns None now and changed the unit test accordingly

test/test_translator.py
abinashkonersman: i changed the name test_get_road_direction_empty_coordinates(): and did break out into appropriate named methods

translator/GCP_cloud_function/cloud_function/main.py
abinashkonersman: I changed it to RuntimeError

test/test_translator.py
abinashkonersman: I did break down those unit test and renamed the function name with specific direction name

test/test_translator.py
abinashkonersman: I updated the unit test and also checked  assert callback.called and test_feature == None

test/test_translator.py
abinashkonersman: i added another test function  named _test_parse_incident_from_coordinates_success_ which calculates direction based on coordinates.

translator/source_code/icone_translator.py
abinashkonersman: i changed it and rather than raise an exception,I logged a warning, and return False.

_**_if not direction:
                logging.warning(f'Invalid incident with id = {incident.get("@id")}.unable to parse direction from street name or polyline')
                return False_**_

test/test_translator.py
abinashkonersman: > Can we split `test_get_event_status()` out to it's three separate units of test? This is a great show of testing the full functionality in that method, but we'll want them to be separate tests.

I splitted into separate unit test

test/test_translator.py
abinashkonersman: i removed it

test/test_translator.py
abinashkonersman: i renamed it to test_validate_incident_invalid

test/test_main.py
abinashkonersman: i removed that duplicated method

test/test_translator.py
payneBrandon: I still see this method in the code.

test/test_main.py
```diff
@@ -0,0 +1,130 @@
...
+    actual_output= main.parse_xml(test_parse_xml_string)
+    assert actual_output == test_valid_output
+
+
+@patch('google.cloud.secretmanager.SecretManagerServiceClient')
+def test_get_ftp_credentials(secret):
+    os.environ['icone_ftp_username_secret_name']='secret_username'
+    os.environ['icone_ftp_password_secret_name']='secret_password'
+    os.environ['project_id'] = 'project_id'
+    main.get_ftp_credentials()
```
payneBrandon: This method returns a username and password. Let's verify those are as expected.

test/test_translator.py
payneBrandon: This is still missing edge cases such as no/invalid info object, and callback. We're also missing testing logic for missing `datafeed_frequency_update`, no incidents found, etc.

test/test_translator.py
```diff
@@ -138,8 +209,78 @@ def test_parse_polyline() :
...
+    assert  test_coordinates == expected_coordinates
 
-def test_get_road_direction():
+def test_parse_polyline_invalid_data() :
+    test_polyline= 'invalid' 
+    test_coordinates=icone_translator.parse_polyline(test_polyline)
+    expected_coordinates= []
+    assert  test_coordinates == expected_coordinates
+
+class TestTranslator(unittest.TestCase):
```
payneBrandon: If we want to use the class, this should be a wrapper for all the methods in this file, not just the individual test.

test/test_translator.py
```diff
@@ -138,8 +209,78 @@ def test_parse_polyline() :
...
+  def test_parse_polyline_invalid_coordinates(self):
+    test_polyline = 'a,b,c,d'
+    with pytest.raises(RuntimeError) as runtimeErr:
```
payneBrandon: We've passed `self` as a parameter to this method and then don't use it. If we just want to use teh `pytest.raises(xxx)` method, there's no need for the class usage and injecting `self`.

translator/source_code/icone_translator.py
payneBrandon: I'm not seeing this change, latest on dev branch still shows returning an empty array here.

test/test_translator.py
abinashkonersman: i added few edge cases and merged it to dev.

translator/source_code/icone_translator.py
abinashkonersman: it should be there now.i fixed it

translator/source_code/icone_translator.py
abinashkonersman: i did it and pushed the updated code to dev

test/test_main.py
abinashkonersman: i added the edge case and pushed the updated code to dev

test/test_main.py
abinashkonersman: i splitted the unit test and updated the edge cases and pushed it to dev.

test/test_main.py
abinashkonersman: i verified and updated the code.should be fixed

test/test_main.py
abinashkonersman: The function does not have a return value and i removed the variable assignment.

test/test_translator.py
abinashkonersman: i added some edge cases according to your comment..and updated to dev

test/test_translator.py
abinashkonersman: Added test cases for invalid info objects : test_wzdx_creator_invalid_info_object
no incident found : test_wzdx_creator_no_incidents()
ni info object : test_wzdx_creator_no_info_object()

test/test_main.py
abinashkonersman: the function is updated to not throw an exception.

test/test_main.py
abinashkonersman: i renamed and updated the function.should be fixed

test/test_main.py
abinashkonersman: i changed the function.should be fixed

translator/GCP_cloud_function/cloud_function/main.py
abinashkonersman: i added another 2 test cases and pushed it to dev

test/test_translator.py
abinashkonersman: i think i separeted it and named it 
1.test_get_road_direction_null_coordinates()
2.test_get_road_direction_empty_coordinates()
3.test_get_road_direction_empty_string():

test/test_translator.py
abinashkonersman: i updated the test function.It should be fixed the problem

test/test_translator.py
```diff
@@ -195,110 +354,253 @@ def test_get_road_direction():
...
+    'starttime': '2020-12-07T14:18:00Z',
+    'creationtime': '2020-12-13T14:18:00Z', 
+    'description': 'Road constructions are going on',
+    'location': {
+      'polyline': '34.8380671,-114.1450650,34.8380671,-114.1450650',
+      'street': 'I-70 N'
+    }
+  }]}}
+
+  #[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z
```
payneBrandon: Can we remove?

test/test_translator.py
```diff
@@ -195,110 +354,253 @@ def test_get_road_direction():
...
+    'starttime': '2020-12-07T14:18:00Z',
+    'creationtime': '2020-12-13T14:18:00Z', 
+    'description': 'Road constructions are going on',
+    'location': {
+      'polyline': '34.8380671,-114.1450650,34.8380671,-114.1450650',
+      'street': 'I-70 N'
+    }
+  }]}}
+  
+#[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z
```
payneBrandon: This seems to have been copy/pasted maybe? Can we clean up this file and remove the commented sections that don't make sense?

abinashkonersman: removed the commented section and pushed it to dev

test/test_translator.py
abinashkonersman: those are regular expression i used i did put it in comment. I just removed those .should be fixed

test/test_translator.py
abinashkonersman: i changed this function .should be fixed now.

translator/source_code/icone_translator.py
```diff
@@ -46,14 +57,17 @@ def wzdx_creator(messages, info):
...
 
-        feature = parse_incident(incident)
+    for incident in messages['incidents']['incident']:
+        # Parse Incident to WZDx Feature    
+        feature = parse_incident(incident, callback_function=unsupported_message_callback)
         if feature:
             wzd['features'].append(feature)
-
+    if not wzd['features']:
+        return None
```
payneBrandon, line 72: Is this the correct logic here? We're saying that if we find any incidents with invalid features we don't generate any WZDx messages. So we might have 50 incidents and if a single one fails the full ingest doesn't succeed. Is this right?

jacob6838: wzd['features'] is the compiled list of features, checked after the for loop which populates it. If it is completely empty (not a single incident), then it will return None instead of returning a complete WZDx object with an empty list of features

payneBrandon: Ohh I got it, that makes sense then!

test/test_translator.py
payneBrandon: This looks good, the only case I still see to test is passing `-h` to cause the `sys.exit()` to be called.

test/test_translator.py
```diff
@@ -195,110 +354,253 @@ def test_get_road_direction():
...
+  
+def test_wzdx_creator_invalid_incidents_no_description() :
+  icone_obj = {'incidents': {'incident': [{
+    '@id': 'U13631595_202012160845',
+    'updatetime': '2020-12-16T17:18:00Z',
+    'starttime': '2020-12-07T14:18:00Z',
+    'creationtime': '2020-12-13T14:18:00Z', 
+    'location': {
+      'polyline': '34.8380671,-114.1450650,34.8380671,-114.1450650',
+      'street': 'I-70 N'
```
payneBrandon, line 449: Can we add to these tests to show that we can have a set of data with both good and bad incidents and we still get the valid incidents back?

test/test_translator.py
payneBrandon: @abinashkonersman Still seeing this dead function, can we remove?

translator/source_code/icone_translator.py
payneBrandon: Still not fixed: 
image

translator/source_code/icone_translator.py
```diff
@@ -353,6 +391,44 @@ def parse_incident(incident):
...
+    if not direction:
+        direction=get_road_direction(coords)
+        if not direction:
+            logging.warning(f'Invalid incident with id = {incident.get("@id")}.unable to parse direction from street name or polyline')
+            return False
+    required_fields = [location, polyline, coords, street, starttime, description, creationtime, updatetime, direction]
+    for field in required_fields:
+        if not field:
+            return False
+            logging.warning(f'Invalid incident with id = {incident.get("@id")}. Not all required fields are present')
```
payneBrandon, line 425: This logging statement never gets hit, can we move above the return line?

translator/source_code/icone_translator.py
```diff
@@ -353,6 +391,44 @@ def parse_incident(incident):
...
 
+#function to validate the incident
+def validate_incident(incident):
+    
+    if not incident or (type(incident) != dict and type(incident) != OrderedDict):
+        logging.warning('incident is empty or has invalid type')
+        return False
+    
+    location = incident.get('location', {})
+    polyline = location.get('polyline')
```
payneBrandon, line 408: If we fail to get location from incident, this line will cause an exception. We check for location existence below so I assume this is not what we want.

jacob6838: The second parameter to dictionary.get is what to return if the key does not exist, and for location we set the return value to an empty dictionary. Then, when we access location.get('polyline'), polyline will be None because 'polyline' is not present in the empty dictionary

payneBrandon: Ok that makes more sense. With this though, will we ever get anything out of checking the `if not location` below? This will always have a value (empty dictionary) right? If we run into the case where we don't find a location we then continue to grab all the other dictionary values before running into the empty polyline and returning false. It would make more sense not return the empty dictionary here and instead check for the null value before continuing to avoid the extra work.

jacob6838: I just wanted to not have 30 if not *field* statements, I wanted to have 1 check at the end, if it was possible. Certainly there is no reason to check location, since polyline will always be None if location does not exist

test/test_translator.py
abinashkonersman: I added an another test case where we are checking both valid and invalid incidents and it returns wzdx message with only the valid incidents.

translator/source_code/icone_translator.py
abinashkonersman: I moved the logging statement above the _return False_ line.

translator/source_code/icone_translator.py
payneBrandon: One check at the end certainly makes sense with the majority of these fields. Just the ones that are relied on by other fields could be checked separately, which here is really only location.

test/test_main.py
payneBrandon: @abinashkonersman We seem to still be missing this.

test/test_main.py
payneBrandon: This is still not using the unit testing tools to verify the exception.

translator/GCP_cloud_function/cloud_function/main.py
```diff
@@ -42,28 +47,73 @@ def translate_newest_icone_to_wzdx(event, context):
...
   secret_client = secretmanager.SecretManagerServiceClient()
-  username_secret_name = os.environ['icone_ftp_username_secret_name']
-  password_secret_name = os.environ['icone_ftp_password_secret_name']
-  project_id = os.environ['project_id']
+  username_secret_name = os.environ.get('icone_ftp_username_secret_name')
+  password_secret_name = os.environ.get('icone_ftp_password_secret_name')
+  project_id = os.environ.get('project_id')
+           
+  if not username_secret_name or not password_secret_name or not project_id:
+    return None
```
payneBrandon: The rest of this method returns two values. Let's update this return to also return two in order to keep things consistent.

test/test_main.py
```diff
@@ -0,0 +1,199 @@
...
+    secret().access_secret_version = fake_secret_client
+    actual = main.get_ftp_credentials()
+    assert actual == None
+
+@patch.dict(os.environ, {
+    'icone_ftp_username_secret_name': 'fail',
+    'icone_ftp_password_secret_name': 'secret_password',
+    'project_id': 'project_id'})
+@patch('google.cloud.secretmanager.SecretManagerServiceClient')
+def test_get_ftp_secrets_does_not_exist(secret):
```
payneBrandon: This tests the case that the "username" secret doesn't exist. Can we name it such to clarify? Also, since we have it handled separately, can we add a test for missing "password" secret?

translator/GCP_cloud_function/cloud_function/main.py
```diff
@@ -42,28 +47,73 @@ def translate_newest_icone_to_wzdx(event, context):
...
+  project_id = os.environ.get('project_id')
+  unsupported_messages_topic_id = os.environ.get('unsupported_messages_topic_id')
+  if not project_id or not unsupported_messages_callback:
+    return False
+  #update this so that this code will execute once, not everytime the function is called
+  publisher = pubsub_v1.PublisherClient()
+  topic_path = publisher.topic_path(project_id, unsupported_messages_topic_id)
+  #publish unsupported messages into pub/sub topic
+  try:
+    future=publisher.publish(topic_path,str.encode(json.dumps(message, indent=2)),origin='auto_icone_translator_ftp cloud function')
```
payneBrandon: Were we seeing issues with publishing JSON formatted code here? That seems to be the only difference between here and our except line below.

translator/GCP_cloud_function/cloud_function/main.py
```diff
@@ -42,28 +47,73 @@ def translate_newest_icone_to_wzdx(event, context):
...
 
-  return username,password
\ No newline at end of file
+  return username,password
+
+
+def unsupported_messages_callback(message):
+  project_id = os.environ.get('project_id')
+  unsupported_messages_topic_id = os.environ.get('unsupported_messages_topic_id')
+  if not project_id or not unsupported_messages_callback:
```
payneBrandon: Since we're already in the `unsupported_messages_callback` method, we don't need to check for it's existence here.

test/test_main.py
```diff
@@ -0,0 +1,199 @@
...
+def test_unsupported_messages_callback_string(pubsub):
+
+    output=main.unsupported_messages_callback('unsupported_messages')
+    publisher=pubsub().publish
+    publisher.assert_called_with(pubsub().topic_path('project_id', 'unsupported_messages_topic_id'),str.encode(json.dumps('unsupported_messages')),origin='auto_icone_translator_ftp cloud function')
+
+
+@patch('google.cloud.pubsub_v1.PublisherClient')
+@patch.dict(os.environ, {})
+def test_unsupported_messages_callback_invalid(pubsub):
```
payneBrandon: This method is really testing the lack of environment variable. Can we rename it such?

test/test_main.py
abinashkonersman: added a test case for password and renamed the other test case to test_get_ftp_secrets_username_does_not_exist

translator/GCP_cloud_function/cloud_function/main.py
abinashkonersman: should be fixed

translator/GCP_cloud_function/cloud_function/main.py
jacob6838: json.dumps can occasionally fail, I think it only happens when it is called with something that is not a string or dict, but just to be safe I had Abinash add a try except to ensure it still publishes no matter what comes in. Could probably replace the try except block with a type check, but it is a rare case anyways so shouldn't make much of a difference

test/test_main.py
abinashkonersman: i renamed it and pushed it to dev

translator/GCP_cloud_function/cloud_function/main.py
abinashkonersman: i changed the return values and updated the unit test

test/test_main.py
```diff
@@ -0,0 +1,208 @@
...
+    secret().access_secret_version = fake_secret_client
+    actual = main.get_ftp_credentials()
+    assert actual == (None, None) 
+
+@patch.dict(os.environ, {
+    'icone_ftp_username_secret_name': 'secret_username',
+    'icone_ftp_password_secret_name': 'fail',
+    'project_id': 'project_id'})
+@patch('google.cloud.secretmanager.SecretManagerServiceClient')
+def test_get_ftp_secrets_username_password_does_not_exist(secret):
```
payneBrandon: We're not testing "username" in here, can we remove from the function name?

abinashkonersman: i removed it

translator/GCP_cloud_function/cloud_function/main.py
payneBrandon: What do you think about refactoring to move the `json.dumps` portion out? Something like:

```
def formatMessage(message):
  try:
    msg = json.dumps(message, indent=2)  
  except:
    msg = message  
  return msg
```

Then you could call the new method from the `unsuppoted_messages_callback` function:
```
future = publisher.publish(topic_path, str.encode(formatMessage(message)))
```

translator/GCP_cloud_function/cloud_function/main.py
payneBrandon: This method is still the same, and still checking for the existence of the method from which you are calling.

jacob6838: It looks like he accidentally had unsupported_messages_callback instead unsupported_messages_topic_id in the if statement. I had him change that, to check if the topic_id exists instead of accidentally checking if the function exists

translator/source_code/icone_translator.py
abinashkonersman: I changed the return type to None and changed the unit test accordingly

abinashkonersman:
> test_get_vehicle_impact

i added an unit test edge case : test_get_vehicle_impact_all_lanes_open

test/test_main.py
abinashkonersman: i added another 2 edge cases for no ftp_url  and invalid ftp_url  which handles the exception case

test/test_translator.py
abinashkonersman: i removed the dead function

test/test_main.py
abinashkonersman: i used the unit test tools and verified the function  and updated the unit test

test/test_translator.py
abinashkonersman: i added the unit test for help (-h)

translator/GCP_cloud_function/cloud_function/main.py
abinashkonersman: i updated the unit test according to your instruction

translator/source_code/icone_translator.py
abinashkonersman: I added a check for location right after retrieving it.

test/test_main.py
payneBrandon: Both new functions don't actually call out to the function to test. They just set it up.

translator/GCP_cloud_function/cloud_function/main.py
payneBrandon: Looks good! Can we remove the commented 2nd call to publish?

abinashkonersman: i removed the commented 2nd call

test/test_main.py
abinashkonersman: I called  out to the function i.e main.translate_newest_icone_to_wzdx(None,None) to test

payneBrandon (approved):
Looks good!

PR #25, opened 2021-04-28
Cv 96

abinashkonersman:
- Added combine_wzdx.py for integration of iCone and COtrip wzdx message
- Added Unit test for combine_wzdx.py
- Updated README.md


payneBrandon (commented):
This PR seems to have much the same changes as in the other PR. Holding off on reviewing until the first PR is merged and we can see the actual changes for this branch.

test/test_combine_wzdx.py
```diff
@@ -0,0 +1,215 @@
+from os import path
+from translator.source_code import combine_wzdx
+from shapely.geometry.polygon import Polygon
```
PTRFRLL, line 3: Looks like we're missing `shapely` in the requirements.txt file. When I attempt to run the unit tests, I get:

```
No module named 'shapely'
```

PTRFRLL (changes requested):

test/test_combine_wzdx.py
abinashkonersman: should be fixed and also added some of the other in requirement.txt

translator/source_code/combine_wzdx.py
```diff
@@ -0,0 +1,157 @@
...
+-c, --cotrip                specify the WZDx cotrip file to compare/combine
+-o, --output                specify the output file for generated wzdx geojson message """
+
+
+def parse_arguments(argv, default_output_file_name='combined_wzdx_message.geojson'):
+    icone = None
+    cotrip = None
+    outputfile = default_output_file_name
+
+    try:
```
PTRFRLL, line 276: Can we display the `help_string` when no arguments are supplied, like the icone/cotrip script? Right now if I run 
```sh
python .\combine_wzdx.py
```
I get an error. Running:

```sh
python .\combine_wzdx.py -i 
```
yields no output

translator/source_code/combine_wzdx.py
```diff
@@ -0,0 +1,157 @@
...
+def combine_wzdx(wzdx_cotrip, wzdx_icone, icone_feature):
+    combined_wzdx = copy.deepcopy(wzdx_cotrip)
+    combined_wzdx['features'][0]['properties']['vehicle_impact'] = icone_feature['properties']['vehicle_impact']
+    combined_wzdx['road_event_feed_info']['data_sources'].append(
+        wzdx_icone['road_event_feed_info']['data_sources'][0])
+
+    return combined_wzdx
+
+
+def iterate_feature(polygon, wzdx_message):
```
PTRFRLL: Is it possible for multiple `Features` to contain points that are in the defined polygon? Looks like this returns the first Feature, is there ever a case where we would have more than one?

translator/source_code/combine_wzdx.py
```diff
@@ -0,0 +1,157 @@
...
+
+        else:
+            print('no duplicate messages were found')
+
+
+def combine_wzdx(wzdx_cotrip, wzdx_icone, icone_feature):
+    combined_wzdx = copy.deepcopy(wzdx_cotrip)
+    combined_wzdx['features'][0]['properties']['vehicle_impact'] = icone_feature['properties']['vehicle_impact']
+    combined_wzdx['road_event_feed_info']['data_sources'].append(
+        wzdx_icone['road_event_feed_info']['data_sources'][0])
```
PTRFRLL: Would it be possible for `wzdx_icone['road_event_feed_info']['data_sources']` to be an empty list? Do we need to check that the `data_sources` property contains an element?

translator/source_code/combine_wzdx.py
```diff
@@ -0,0 +1,157 @@
...
+import sys
+
+
+def main():
+
+    icone_file, cotrip_file, outputfile = parse_arguments(
+        sys.argv[1:], default_output_file_name='combined_wzdx_message.geojson')
+
+    with open(outputfile, 'w+') as f:
+        wzdx_icone = json.loads(open(
```
PTRFRLL, line 44: We may want to handle the exception that occurs when a non-JSON file is supplied.

translator/source_code/combine_wzdx.py
```diff
@@ -0,0 +1,157 @@
...
+
+def iterate_feature(polygon, wzdx_message):
+    for feature in wzdx_message['features']:
+        for coord in feature['geometry']['coordinates']:
+            if isPointInPolygon(Point(coord[1], coord[0]), polygon):
+                return feature
+
+
+# generate polygon from list of geometry ([[long, lat], ...]) and width in meters
+def generate_polygon(geometry, polygon_width):
```
PTRFRLL: Can we make this variable name more descriptive?
```suggestion
def generate_polygon(geometry, polygon_width_in_meters):
```

translator/source_code/combine_wzdx.py
```diff
@@ -0,0 +1,157 @@
...
+        sys.argv[1:], default_output_file_name='combined_wzdx_message.geojson')
+
+    with open(outputfile, 'w+') as f:
+        wzdx_icone = json.loads(open(
+            icone_file, 'r').read())
+        wzdx_cotrip = json.loads(open(
+            cotrip_file, 'r').read())
+
+        polygon = generate_polygon(
+            wzdx_cotrip['features'][0]['geometry']['coordinates'], 100)
```
PTRFRLL: Why was 100 chosen? Can we add it to a variable so its not a "magic number"?

```suggestion
            polygon_width_in_meters = 100
            wzdx_cotrip['features'][0]['geometry']['coordinates'], polygon_width_in_meters)
```

PTRFRLL (changes requested):

test/test_combine_wzdx.py
payneBrandon: Unit tests are still failing to run:
image

README.md
```diff
@@ -22,8 +20,7 @@ Requires:
   - xmltodict
```
payneBrandon: Now that we have significantly more requirements here, can we update this to just say we need Python 3.6 or higher, and packages as specified in the `requirements.txt`?

README.md
```diff
@@ -71,27 +69,41 @@ python -m translator.source_code.icone_translator -i 'translator/sample files/ic
...
 Example usage:
+
 ```
 python -m translator.source_code.cotrip_translator -i 'translator/sample files/cotrip_data/cotrip_1.json' 
 ```
 
-### Unit Testing
+### Execution for Combine_wzdx
+
+#### Step 1: Run the translator script (from Work_Zone/translator/source_code)
```
payneBrandon: Since there's only a single step, I don't think we need to call out "step 1..."

translator/source_code/combine_wzdx.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def iterate_feature(polygon, wzdx_message):
+    for feature in wzdx_message['features']:
+        for coord in feature['geometry']['coordinates']:
+            if isPointInPolygon(Point(coord[1], coord[0]), polygon):
+                return feature
+
+
+# generate polygon from list of geometry ([[long, lat], ...]) and width in meters
```
payneBrandon: What sort of a polygon are we creating, and for what purpose? Are we making a bounding box, a line string? Can we be more descriptive here so we know what we are doing and why?

payneBrandon: Also, if we move these comments inside the method and use triple quotations they will show up on calls to the method. For example: 
```
def generate_polygon(geometry, polygon_width):
    """generate polygon from list of geometry ([[long, lat], ...]) and width in meters
    
    Args: 
        geometry: something
        polygon_width: width in meters
    """
```

When hovering over calls to this method, now we get a nice description:
image

translator/source_code/combine_wzdx.py
```diff
@@ -0,0 +1,157 @@
...
+        return None
+    geodesic_pyproj = pyproj.Geod(ellps='WGS84')
+
+    # Initializing lists to create polygon
+    polygon_left_points = []
+    polygon_right_points = []
+
+    for i in range(0, len(geometry)):
+        # Set up points to calculate heading
+        if i == 0:
```
payneBrandon, line 194: This seems like it would repeat the first points (position 0, and 1) the second time through. Can we remove that repeat?

translator/source_code/combine_wzdx.py
```diff
@@ -0,0 +1,157 @@
...
+            # first point, heading from first to second point
+            p1 = geometry[i]
+            p2 = geometry[i+1]
```
payneBrandon, line 197: If we had a list with only a single element in it this would fail.

translator/source_code/combine_wzdx.py
```diff
@@ -0,0 +1,157 @@
...
+        left_point = geodesic(meters=polygon_width/2).destination(origin, left)
+        right_point = geodesic(meters=polygon_width /
+                               2).destination(origin, right)
+
+        # Append points to left and right lists
+        polygon_left_points.append([left_point.latitude, left_point.longitude])
+        polygon_right_points.append(
+            [right_point.latitude, right_point.longitude])
+
+    # Create list of points in correct order (all left points, then all right points in reverese order)
```
payneBrandon: `reverse` rather that `reverese`

README.md
```diff
@@ -1,12 +1,10 @@
 # Work_Zone
```
payneBrandon: We really have three separate functionalities here - one parsing WZDX messages out of iCone data, one doing the same out of COtrip data, and finally a script to combine those two separate message sets. Since we've upped the complexity can we revamp this README a bit to reflect that? For instance we really only have "how to run the script" type of descriptions here but it would be very helpful to know what business logic we are relying on to do the parsing of the various data sources. 

With that same vein of thought, we should really lay out what business logic the `combine_wzdx.py` is working with. The script itself is not very descriptive on what is being done and why. Let's work a bit on updating this README to explain how we are deciding what messages will be combined, what data we are looking to combine within those message sets, and also what data we are still missing from that combination.

translator/source_code/combine_wzdx.py
```diff
@@ -0,0 +1,157 @@
...
+    combined_wzdx['road_event_feed_info']['data_sources'].append(
+        wzdx_icone['road_event_feed_info']['data_sources'][0])
+
+    return combined_wzdx
+
+
+def iterate_feature(polygon, wzdx_message):
+    for feature in wzdx_message['features']:
+        for coord in feature['geometry']['coordinates']:
+            if isPointInPolygon(Point(coord[1], coord[0]), polygon):
```
payneBrandon, line 170: Are we saying here that any overlap of the two work zones means it is the same work zone and should be combined? Is there ever a chance of separate work zones that happen to overlap a part of geometry?

translator/source_code/combine_wzdx.py
```diff
@@ -0,0 +1,157 @@
...
+            cotrip_file, 'r').read())
+
+        polygon = generate_polygon(
+            wzdx_cotrip['features'][0]['geometry']['coordinates'], 100)
+        feature = iterate_feature(polygon, wzdx_icone)
+        if feature:
+
+            f.write(json.dumps(combine_wzdx(
+                wzdx_cotrip, wzdx_icone, feature), indent=2))
+            print('duplicate iCone and COtrip messages were found and combined WZDx message was written to combined_wzdx_message.geojson file')
```
payneBrandon: "Duplicate" is not quite the right terminology here, we have overlapping datasets that correspond to the same work zone and provide different data that we are combining into a single new source. Can we update these print messages to be a bit more descriptive with that in mind? Maybe just "Combined iCone and COtrip messages written to xxx".

translator/source_code/combine_wzdx.py
```diff
@@ -0,0 +1,157 @@
...
+
+        else:
+            print('no duplicate messages were found')
+
+
+def combine_wzdx(wzdx_cotrip, wzdx_icone, icone_feature):
+    combined_wzdx = copy.deepcopy(wzdx_cotrip)
+    combined_wzdx['features'][0]['properties']['vehicle_impact'] = icone_feature['properties']['vehicle_impact']
```
payneBrandon: Is this `vehicle_impact` value guaranteed to be in both iCone and COtrip data?

translator/source_code/combine_wzdx.py
```diff
@@ -0,0 +1,157 @@
...
+
+            f.write(json.dumps(combine_wzdx(
+                wzdx_cotrip, wzdx_icone, feature), indent=2))
+            print('duplicate iCone and COtrip messages were found and combined WZDx message was written to combined_wzdx_message.geojson file')
+
+        else:
+            print('no duplicate messages were found')
+
+
+def combine_wzdx(wzdx_cotrip, wzdx_icone, icone_feature):
```
payneBrandon: With this combine function, it looks like we are just taking everything from COtrip, then adding a single feature `vehicle_impact` value. That seems limited for a combined stream, is it possible there's more data in iCone than we are taking here? What about differing geometry for instance - could we have iCone with multiple points outside the polygon defined above or are we always getting a single point with iCone?

translator/source_code/combine_wzdx.py
abinashkonersman: should be fixed and i changed the name.

README.md
abinashkonersman: i specified requirements.txt and changed it. it should be fixed now

README.md
abinashkonersman: should be fixed..i only did just put the word "step" instead of "step 1"

translator/source_code/combine_wzdx.py
abinashkonersman: should be fixed

translator/source_code/combine_wzdx.py
abinashkonersman: should be fixed

translator/source_code/combine_wzdx.py
payneBrandon: @abinashkonersman What I mean here is we are assigning the `vehicle_impact` value from the `icone_feature` object. Is that object guaranteed to have a "properties" property with a "vehicle_impact" property or is it possible for that property to be missing?

Likewise, we're looking to the first feature in the `combined_wzdx` object. Is that object guaranteed to have a list of features, each with "properties" and "vehicle_impact" properties? We likely need to add a check here to ensure this data structure and object properties exist  before assigning to them to prevent any issues in the future.

translator/source_code/combine_wzdx.py
abinashkonersman: should be fixed

translator/source_code/combine_wzdx.py
abinashkonersman: Added a try catch and an exception message to handle the exception that occurs when a non-JSON file is supplied.

translator/source_code/combine_wzdx.py
jacob6838: No, in the WZDx messages that we generate, they each only have one feature

translator/source_code/combine_wzdx.py
jacob6838: The WZDx message spec requires that there be at least one data source. However this should probably append all data sources, not just the first one

translator/source_code/combine_wzdx.py
abinashkonersman: should be fixed

translator/source_code/combine_wzdx.py
jacob6838: This is just to calculate the instantaneous heading, The first point uses the forward heading, the other points use the reverse heading, it is impossible to get a heading for every point without repeating. In line 72 p1 gets reset to geometry[i]

translator/source_code/combine_wzdx.py
jacob6838: Line 44 ensures that the geometry is at least >=1 in length

payneBrandon: Ah I see, we're checking if the list length <= 1 then returning. This ensures the geometry length is > 1 (>= would cause issues).

translator/source_code/combine_wzdx.py
payneBrandon: Couldn't we just use the one set of logic grabbing the next item in the array and go one index less to get rid of duplicate geometries? 
```
for i in range(0, len(geometry) -1)
     current_pt = geometry[i]
     next_pt = geometry[i+1]

     # Get forward heading between 2 points
     fwd_heading, _, __ = geodesic_pyproj.inv(
          current_pt[0], current_pt[1], next_pt[0], next_pt[1])
```

Then we don't need to reset the "pt1" either, just use "current_pt" for the rest below.

jacob6838: The issue comes in when we reach the end of geometry, and we need to generate a heading for the final point. In this case you would look at the heading for the previous point and current point, which would add the second case

translator/source_code/combine_wzdx.py
jacob6838: The iCone message really doesn't have a ton of information, compared to the cotrip/salesforce message. There are only a few fields that are more informative in iCone than in COtrip, and this is (currently) just the vehicle impact

jacob6838: I believe we are iterating over the iCone geometry to catch any points that are in the polygon, on line 36

translator/source_code/combine_wzdx.py
payneBrandon: Ah ok, I see what you're doing here - you're creating a closed path based on a small buffer around the points to generate a Polygon you can then use to find intersecting points. This makes more sense now.

translator/source_code/combine_wzdx.py
```diff
@@ -0,0 +1,177 @@
...
+
+def iterate_feature(polygon, wzdx_message):
+    for feature in wzdx_message['features']:
+        for coord in feature['geometry']['coordinates']:
+            if isPointInPolygon(Point(coord[1], coord[0]), polygon):
+                return feature
+
+
+# generate polygon from list of geometry ([[long, lat], ...]) and width in meters
+def generate_polygon(geometry, polygon_width_in_meters):
```
payneBrandon:
```suggestion
def generate_buffer_polygon_from_linestring(geometry, polygon_width_in_meters):
```

translator/source_code/combine_wzdx.py
```diff
@@ -0,0 +1,177 @@
...
+    """generate polygon from list of geometry ([[long, lat], ...]) as linestring and width in meters
```
payneBrandon:
```suggestion
    """Generate a polygon from a Linestring using polygon_width_in_meters as a buffer.
```

translator/source_code/combine_wzdx.py
```diff
@@ -0,0 +1,177 @@
...
+
+    Args: 
+        geometry: Linestring
+        polygon_width: width in meters
```
payneBrandon:
```suggestion
        polygon_width_in_meters: width in meters
```
Update to match actual argument

README.md
payneBrandon: What I mean here is that in the `requirements.txt` file we  have  many more libraries than are mentioned here. Instead of listing every one in the README, let's just update this README file to say Python 3.6 or higher is required, as well as several supporting libraries as seen in requirements.txt.

README.md
payneBrandon:
```suggestion
#### Run the translator script (from Work_Zone/translator/source_code)
```

README.md
payneBrandon: @abinashkonersman any progress on this one?

translator/source_code/combine_wzdx.py
payneBrandon: Line numbers are a bit screwy now with all the changes going on...are you meaning the `iterate_feature` method? It looks like we take in the polygon generated from the COtrip data an an iCone message. Then we iterate over the features in iCone, look to the coordinates, and determine if any coordinates are in the polygon. We then return the first feature we find with a single point contained in the polygon. This could leave out additional features (with the additional vehicle impacts), or points not in the polygon if any such points exist.

translator/source_code/combine_wzdx.py
abinashkonersman: changed it to polygon_width_in_meters

translator/source_code/combine_wzdx.py
abinashkonersman: @payneBrandon should be fixed

README.md
abinashkonersman: i thought i changed it but i guess i did not ..it should be fixed now

translator/source_code/combine_wzdx.py
abinashkonersman: changed the name to generate_buffer_polygon_from_linestring

README.md
abinashkonersman: i changed it there

translator/source_code/combine_wzdx.py
jacob6838: Our intent was to make a basic algorithm that would work most of the time. I have had a few discussions with tony where we came up with tons of different algorithms, including heading checks, checking whether the iCone linestring crosses any point of the polygon, matching both to roads using a maps api, and even much simpler options like using a rectangle instead of a polygon for increased performance. 
**I would not at all be opposed to implementing a heading check**, I just don't want to spend too much time perfecting this algorithm when we don't have any examples of duplicated workzones. I say that after we already wrote the polygon generator, sunk cost I guess? who knows haha

README.md
abinashkonersman: @payneBrandon i added that description after Google Cloud Function description and diagram.Please check and let me know if there is any comment.

translator/source_code/combine_wzdx.py
jacob6838: Each **feature** is required to have properties, and **properties** is required to have **vehicle_impact**. It is however theoretically possible for a WZDx message to have an empty set of features, this should be corrected.

translator/source_code/combine_wzdx.py
jacob6838: ok I see what you are saying now. You are correct, we stop looking after a single feature matches. **We made the assumption that duplicated workzones would not appear in a single iCone dataset**, I think that is a valid assumption. However, when this gets expanded to more datasets, this could potentially be a problem. I will leave it up to @payneBrandon to decide whether it is worth resolving

translator/source_code/combine_wzdx.py
payneBrandon: @abinashkonersman  Unit tests will likewise need to be updated for the new name. Currently they are failing.

abinashkonersman: @payneBrandon updated the unit test.. all are passing now.

payneBrandon:
@abinashkonersman with the approval of #26 there are some overlapping changes, including those in the README. Can we get the latest dev branch changes merged into this branch so we're just dealing with changes in this PR?

translator/source_code/combine_wzdx.py
abinashkonersman: @PTRFRLL  should be fixed

translator/source_code/combine_wzdx.py
abinashkonersman: @payneBrandon should be fixed

translator/source_code/combine_wzdx.py
abinashkonersman: @payneBrandon  should be fixed

README.md
payneBrandon: Can we still remove the "step: " text? It reads quite awkward.

README.md
```diff
@@ -89,6 +104,10 @@ A system was created in google cloud platform to automatically translate iCone d
 
 ![alt text](translator/GCP_cloud_function/iCone%20Translator%20block%20diagram.png)
 
+### Message Combination Logic:
+
+The combined WZDx script file combines each set of duplicated messages from iCone and COtrip to a single improved WZDx message.It begins with COtrip message and adds data from iCone and publishes a combined WZDx message as an output.The script file checks for the vehicle_impact from iCone WZDx message and adds it    to the Combined WZDx output message file along with all  data sources present in  iCone message.We intend to add more fields in future but currently only vehicle_impact and datasources are being combined.
```
payneBrandon:
```suggestion
The `combine_wzdx` script file combines the output from the iCone and COtrip translators, based on overlapping geography, into a single improved WZDx message. The COtrip message set contains significantly more data, and is used as the base for this new combined message. The script then finds any geographically co-located messages from the iCone data set, pulls in the additional information (comprised of vehicle impact data and data sources) and publishes a new, combined WZDx message. Future state of this script will include additional data fields from the iCone data set as they become available.
```

README.md
payneBrandon: I added a bit to the combine description - take a look and let me know what you think. Can we similarly add some text talking about the logic to parse out iCone and COtrip data and what fields are currently supported? Ideally we want anyone to be able to pick up this repo, plug in an iCone data feed (or COtrip) and know exactly what the code is doing.

translator/source_code/combine_wzdx.py
```diff
@@ -0,0 +1,299 @@
...
+        wzdx_icone: Destination WZDx message (dict)
+        icone_feature_index: Index of source feature to combine (int)
+
+    Returns: 
+        Destination WZDx message, updated from source if new data was found
+    """
+
+    # If indexes do not exist for either source or destination, immediately return destination object
+    if len(wzdx_destination['features']) <= destination_feature_index or len(wzdx_source['features']) <= source_feature_index:
+        destination_feature = wzdx_destination['features'][destination_feature_index]
```
payneBrandon: Can we remove? This would blow up if the destination_feature_index was greater than the length and we aren't using this variable anywhere. 
```suggestion
```

jacob6838: Completely agree, not sure how that made it through, my mistake

README.md
abinashkonersman: @payneBrandon It should be fixed now.

translator/source_code/combine_wzdx.py
abinashkonersman: @payneBrandon I removed it.

README.md
abinashkonersman: @payneBrandon  I changed the description .

README.md
abinashkonersman: @payneBrandon  we have [Word docs](https://github.com/CDOT-CV/Work_Zone/blob/CV-96/translator/docs/WZDx%20Translator%20Field%20Mappings.docx) that describes all the fields of iCone and Cotrip data feed.

abinashkonersman: @payneBrandon  should be fixed i added a little detailed description.

payneBrandon: I like the link to our documentation that you added in. Next sprint we'll want to improve that documentation and update it a bit (we still reference spec 3 instead of 3.1 in a few places for instance).

payneBrandon (approved):

PR #29, opened 2021-08-24
64753 improve 568 translator

jacob6838:
- Support numbered keys  ("streetNameFrom2", ...)
  - unit tests
- Support all discernable types_of_work from descriptionForProject field


tests/navjoy_translator_test.py
```diff
@@ -1061,3 +1075,614 @@ def test_get_types_of_work_empty():
...
+            'mileMarkerEnd2': None,
+            'mileMarkerStart2': None,
+        }
+    }]
+    actual = navjoy_translator.expand_speed_zone(event)
+
+    assert expected == actual
+
+
+# I hate how long this test is, but this is what is has to be to test all 4 at the same time
```
payneBrandon, line 1158: Not sure it has to be done, but what do you think about extracting these "expected" values to JSON files in a separate directory (maybe just `./ExpectedTestResults`)? It would clean up the tests a bit, we'd just have to read/parse the docs.

translator/navjoy_translator.py
```diff
@@ -67,21 +121,49 @@ def wzdx_creator(messages, info=None, unsupported_message_callback=None):
...
-        # closures can cover both directions
-        directions = message.get('data', {}).get('directionOfTraffic')
 
-        for direction in get_directions_from_string(directions):
-            # Parse closure to WZDx Feature
+        separated_messages = expand_speed_zone(message)
+        for msg in separated_messages:
             feature = parse_reduction_zone(
-                copy.deepcopy(message), direction, callback_function=unsupported_message_callback)
+                copy.deepcopy(msg), callback_function=unsupported_message_callback)
```
payneBrandon: Do we need to `deepcopy` the msg here? We're already working with a copy and don't seem to use the value after this parse method anyway.

jacob6838: I was having some initial pass by reference issues, so I added in deepcopies everywhere. I will take a look and see which ones are actaully necessary

jacob6838: Just removed this

tests/navjoy_translator_test.py
jacob6838: Broke these out into another file, called navjoy_translator_test_expected_results.py

payneBrandon (approved):
Looks good to me!

PR #31, opened 2021-08-31
Set up CI with Azure Pipelines

jacob6838:
Creating initial CI CD GCP function deployment pipeline

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,27 @@
+# Starter pipeline
+# Start with a minimal pipeline that you can customize to build and deploy your code.
+# Add steps that build, run tests, deploy, and more:
+# https://aka.ms/yaml
+
+trigger:
+- 64097-GCP-hosting
+
+pool:
+  vmImage: ubuntu-latest
+
+steps:
+
+# - script:
+#     gcloud auth activate-service-account --key-file $(authkey.secureFilePath)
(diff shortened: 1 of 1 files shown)


jacob6838:
This is a test very simple pipeline

azure-pipelines.yml
```diff
@@ -0,0 +1,27 @@
...
+#   displayName: 'deploy icone function'
+
+# - script:
+#     gcloud auth activate-service-account --key-file $(authkey.secureFilePath)
+#     gcloud functions deploy salesforce-data --runtime--region=us-central1 --project=cdot-oim-wzdx-prod
+#   displayName: 'deploy cotrip function'
+
+- script:
+    # gcloud auth activate-service-account --key-file $(authkey.secureFilePath)
+    gcloud functions deploy navjoy-568-translator--region=us-central1 --project=cdot-oim-wzdx-dev
```
payneBrandon: Do we already have a build for `navjoy-568-translator` that this just pulls from? Also we probably need a space between `navjoy-568-translator` and the `--region` parameter.

jacob6838: Good point on the space, just fixed that. It is configured to build in GCP, so it will just deploy the whole local folder

payneBrandon (approved):
Looks good to me. Once we have this POC working, we'll want to abstract out the project name so we can use the same pipeline across projects.

PR #35, opened 2021-09-03
Update azure-pipelines.yml for Azure Pipelines

jacob6838:


jacob6838:
This pipeline deploys the iCone and Navjoy cloud functions to GCP. The projects and function names are defined by variables. The pipeline cannot currently access the COTrip translator because it is in the Prod project, and I have not yet created the service account. I am planning to leave that for next sprint, the COTrip translator needs some work. There is currently no data for it to access, so it might be good to take a look into the RTDH and figure out why there is no data coming in

azure-pipelines.yml
```diff
@@ -4,11 +4,20 @@
 # https://aka.ms/yaml
 
 trigger:
-- 64097-GCP-hosting
+- 6666-automatic-gcp-deployment
```
payneBrandon: Will we want to set this trigger to be the main branch?

jacob6838: Eventually yes. For now the best bet is probably dev, maybe we should create a pull to Main soon?

azure-pipelines.yml
```diff
@@ -18,20 +27,19 @@ steps:
     secureFile: 'GoogleServiceAccountKey.json'
```
payneBrandon, line 27: Is this automatically set when we turned on that plugin? Or is this something you added on the DevOps side of things?

jacob6838: This is something I had to upload to DevOps, that is a key that allows the pipeline to use the service account

jacob6838: https://dev.azure.com/SOC-OIT/CDOT/_library?itemType=SecureFiles&view=SecureFileView&secureFileId=99077f4b-bea2-4fe5-ba85-fd33fce50adb&path=GoogleServiceAccountKey.json

azure-pipelines.yml
payneBrandon: Maybe we should have two pipelines, one for dev and one for prod? Then when we merge to dev branch the dev pipeline kicks off and pushes to our dev environment. And the prod one would just trigger off the main  branch and push to our production environment. Thoughts?

azure-pipelines.yml
payneBrandon: I figured it was the service account auth file, but just wasn't sure if this was something we had to do manually or not. Thanks for the clarification!

azure-pipelines.yml
jacob6838: That sounds reasonable. It is a little awkward with COTrip currently because its resources only exist in prod (as far as I know). but generally I think that makes perfect sense

payneBrandon: Ok, for now let's just update this to trigger off dev and push out. Next sprint we can add the production release.

azure-pipelines.yml
```diff
@@ -4,11 +4,20 @@
...
+  cotrip_project_name: 'cdot-oim-wzdx-prod'
+  navjoy_project_name: 'cdot-oim-wzdx-dev'
+
+  icone_function_name: 'auto_icone_translator_ftp'
+  cotrip_function_name: 'salesforce-data'
+  navjoy_function_name: 'navjoy-568-translator'
+
 steps:
 
 - task: DownloadSecureFile@1
```
payneBrandon, line 23: Can we add some description for this file either to the existing README or another documentation page? It would be nice to have some text describing what we're doing in this file, specifically how this GoogleServiceAccountKey.json lives in devops and is injected into the pipeline (and how we could update that if needed).

azure-pipelines.yml
jacob6838: Updated to Dev

azure-pipelines.yml
jacob6838: Added description of deployment pipeline in README

payneBrandon (approved):

PR #40, opened 2022-01-18
Planned events

jacob6838:
Adding Open TMS Planned Events translator (v4.0)


payneBrandon:
@jacob6838 looks like there's a couple of merge conflicts now that we've got that last PR into the dev branch.

jacob6838:
Just fixed all of the merge conflicts, and unit tests are succeeding

standard_568_Form568-44819108-703c-4d6d-ae0d-7dca7319e5b0_1638373609_eastbound.json
```diff
@@ -0,0 +1,64 @@
+{
```
payneBrandon: Do all these JSON files at the root need to be here?

wzdx/sample_files/validation_schema/4/WZDxFeed copy.py
```diff
@@ -0,0 +1,64 @@
+wzdx_v40_schema_string = {
```
payneBrandon: Is this one needed?

payneBrandon:
There seems to be quite a few duplicate files between JSON and .py files referencing a 4.0 (or "_40") feed. Are all of these necessary?

wzdx/tools/wzdx_translator.py
```diff
@@ -129,7 +129,37 @@ def add_ids(message):
     return message
 
 
-def initialize_wzdx_object(info):
+# Add ids to message
+# This function may fail if some optional fields are not present (lanes, types_of_work, relationship, ...)
+def add_ids(message, event_type):
```
payneBrandon: Is this specific to v4.0? Should we name it `add_ids_v4` so we know which version we're working with in the future?

wzdx/tools/wzdx_translator.py
```diff
@@ -178,6 +208,104 @@ def initialize_wzdx_object(info):
     return wzd
 
 
+def initialize_wzdx_object(info):
```
payneBrandon: Should we name this `initialize_wzdx_object_v4` to keep with the "v3" naming above and specify version?

wzdx/standard_to_enhanced/planned_events_translator.py
```diff
@@ -0,0 +1,393 @@
+
+import argparse
+import json
+import logging
+import copy
+from collections import OrderedDict
```
payneBrandon: OrderedDict and path don't seem to be used, can we remove?

wzdx/standard_to_enhanced/planned_events_translator.py
```diff
@@ -0,0 +1,393 @@
...
+
+DEFAULT_PLANNED_EVENTS_FEED_INFO_ID = '49253be7-0c6a-4a65-8113-450f9041f989'
+
+
+def main():
+    input_file, output_file = parse_planned_events_arguments()
+    # Added encoding argument because of weird character at start of incidents.xml file
+
+    planned_events_obj = json.loads(open(input_file, 'r').read())
+    wzdx, event_type = wzdx_creator(planned_events_obj)
```
payneBrandon: The `wzdx_creator` method seems to return a Dict object, but here we're trying to set two variables?

wzdx/standard_to_enhanced/planned_events_translator.py
```diff
@@ -0,0 +1,393 @@
...
+
+    if feature:
+        wzd.get('features', []).append(feature)
+    if not wzd.get('features'):
+        return None
+
+    wzd = wzdx_translator.add_ids(wzd, event_type)
+    return wzd
+
+# {
```
payneBrandon: Do we need to keep this in here or can we reference it as a file?

wzdx/standard_to_enhanced/planned_events_translator.py
```diff
@@ -0,0 +1,393 @@
...
+    for key, value in properties.items():
+        if not value and key not in ['road_event_id', 'data_source_id']:
+            del filtered_properties[key]
+
+    feature = {}
+    feature['type'] = "Feature"
+    feature['properties'] = filtered_properties
+    feature['geometry'] = geometry
+    feature['id'] = event.get('source', {}).get('id', uuid.uuid4())
+    print(feature)
```
payneBrandon: Should we leave this print in?

wzdx/standard_to_enhanced/planned_events_translator.py
```diff
@@ -0,0 +1,393 @@
...
+    properties = {}
+
+    # I included a skeleton of the message, fill out all required fields and as many optional fields as you can. Below is a link to the spec page for a road event
+    # https://github.com/usdot-jpo-ode/jpo-wzdx/blob/master/spec-content/objects/RoadEvent.md
+
+    core_details = {}
+
+    # # id
+    # # Leave this empty, it will be populated by add_ids_v3
+    # core_details['road_event_id'] = ''
```
payneBrandon: Should this be setting the `road_event_id` field to empty, or leaving the whole field out?

wzdx/standard_to_enhanced/planned_events_translator.py
```diff
@@ -0,0 +1,393 @@
...
+    for key, value in properties.items():
+        if not value and key not in ['road_event_id', 'data_source_id', 'end_date']:
+            del filtered_properties[key]
+
+    feature = {}
+    feature['type'] = "Feature"
+    feature['properties'] = filtered_properties
+    feature['geometry'] = geometry
+    feature['id'] = event.get('source', {}).get('id', uuid.uuid4())
+    print(feature)
```
payneBrandon: Do we want to print the feature each time?

tests/standard_to_enhanced/planned_events_translator_test.py
```diff
@@ -0,0 +1,461 @@
...
+from datetime import datetime, timezone
+from unittest.mock import MagicMock, Mock, patch
+
+import time_machine
+from wzdx.standard_to_enhanced import planned_events_translator
+
+# Unit testing code for navjoy_translator.py
+# --------------------------------------------------------------------------------Unit test for parse_incident function--------------------------------------------------------------------------------
+
+# utc_time_zone = ZoneInfo("America/Los_Angeles")
```
payneBrandon:
```suggestion
```

tests/standard_to_enhanced/planned_events_translator_test.py
```diff
@@ -0,0 +1,461 @@
...
+    'contact_name': 'Ashley Nylen',
+    'contact_email': 'ashley.nylen@state.co.us',
+    'issuing_organization': 'CDOT'
+})
+@patch('uuid.uuid4')
+@unittest.mock.patch('wzdx.standard_to_enhanced.navjoy_translator.datetime')
+@unittest.mock.patch('wzdx.tools.wzdx_translator.datetime')
+def test_wzdx_creator(mock_dt, mock_dt_3, mockuuid):
+    init_datetime_mocks([mock_dt, mock_dt_3])
+    uuid.uuid4 = Mock()
```
payneBrandon: If we're resetting the uuid.uuid4 to be a Mock here, do we need to patch it above?

tests/raw_to_standard/planned_events_test.py
```diff
@@ -0,0 +1,222 @@
+from wzdx.raw_to_standard import planned_events
+from tests.raw_to_standard import planned_events_test_expected_results as expected_results
+import uuid
+import json
+from unittest.mock import MagicMock, Mock, patch
```
payneBrandon: Several of these import statements aren't used, can we clean up?

tests/raw_to_standard/planned_events_test.py
```diff
@@ -0,0 +1,222 @@
...
+        "attributes": {}
+    }
+
+    actual = planned_events.expand_event_directions(event)
+
+    assert expected_results.test_expand_speed_zone_2_expected == actual
+
+
+@patch('uuid.uuid4')
+def test_generate_standard_messages_from_string(mockuuid):
```
payneBrandon: We should either inject a patch of the mockuuid and use it, or just mock the uuid itself in the body, but doing both isn't needed.

tests/raw_to_standard/planned_events_test.py
```diff
@@ -0,0 +1,222 @@
...
+    expected = expected_results.test_generate_standard_messages_from_string_expected
+    for i in expected:
+        del i['rtdh_timestamp']
+        del i['event']['source']['last_updated_timestamp']
+    # actual_standard = [dict(x) for x in actual_standard]
+    print(actual_standard)
+    assert actual_standard == expected
+
+
+def test_get_lanes_list():
```
payneBrandon: This test is named the same as the one below it, so only one of them runs. Do we need to run both?

payneBrandon (changes requested):
A few comments.

wzdx/standard_to_enhanced/planned_events_translator.py
jacob6838: removed

tests/standard_to_enhanced/planned_events_translator_test.py
jacob6838: The patch makes it only apply to the current context. I was having an issue where it would never un-mock a method, unless I use the @patch

tests/raw_to_standard/planned_events_test.py
jacob6838: similar to above answer

tests/raw_to_standard/planned_events_test.py
jacob6838: renamed

tests/standard_to_enhanced/planned_events_translator_test.py
payneBrandon: That's fine to continue using the patch, but then we should actually use it. Here we inject a patch, then override the default uuid.uuid4 as a Mock anyway. Instead, we should set the side_effect of the injected patch. Something like:
image

jacob6838: I can't really explain why, but this fails. If I run it with 
```
mockuuid.uuid4 = Mock()
mockuuid.uuid4.side_effect = 'we234de'
```
then it does not mock the call of uuid.uuidv4() in the called code. The only I could get it to work was with uuid.uuidv4

jacob6838: Updated to use 
```
@patch.object(uuid, 'uuid4', side_effect=['w', '', '3'])
```
instead

wzdx/raw_to_standard/planned_events.py
```diff
@@ -311,15 +318,15 @@ def create_rtdh_standard_msg(pd):
                 "end_timestamp": date_tools.date_to_unix(end_date),
             },
             "detail": {
-                "road_name": pd.get("properties/routeName"),
-                "road_number": pd.get("properties/routeName"),
+                "road_name": roadName,
+                "road_number": roadName,
```
payneBrandon, line 346: Is this "road_number" expected to be numeric, and if so will we ever run into issues if we get a road name that isn't?

wzdx/standard_to_enhanced/planned_events_translator.py
```diff
@@ -0,0 +1,302 @@
+
+import argparse
+import json
+import logging
+import copy
+import uuid
+
+from wzdx.sample_files.validation_schema import wzdx_v40_feed, road_restriction_v40_feed
+
+from wzdx.tools import date_tools, polygon_tools, wzdx_translator
```
payneBrandon:
```suggestion
from wzdx.tools import date_tools, wzdx_translator
```
Doesn't look like polygon_tools is being used

wzdx/standard_to_enhanced/planned_events_translator.py
```diff
@@ -0,0 +1,302 @@
...
+
+DEFAULT_PLANNED_EVENTS_FEED_INFO_ID = '49253be7-0c6a-4a65-8113-450f9041f989'
+
+
+def main():
+    input_file, output_file = parse_planned_events_arguments()
+
+    planned_events_obj = json.loads(open(input_file, 'r').read())
+    wzdx = wzdx_creator(planned_events_obj)
+    print(wzdx['road_event_feed_info'])
```
payneBrandon: Rather than "print" should we use the logging.debug and add a bit to it?

wzdx/tools/polygon_tools.py
```diff
@@ -126,6 +126,35 @@ def get_road_direction_from_coordinates(coordinates):
     return direction
 
 
+# function to get road direction by using geometry coordinates
+def get_road_directions_from_coordinates(coordinates):
```
payneBrandon, line 130: Is this direction supposed to line up with the overall labeled road direction? If so, eventually we'll have to revisit this logic. There's some odd places where this wouldn't hold up - for instance if the coordinates happen to be in a bend in the road for shoulder work, or a large section of highway tending northward that is actually on a road labeled for westbound traffic.

payneBrandon (commented):
Just a couple comments, looking good.

wzdx/raw_to_standard/planned_events.py
jacob6838: road_number ideally should be numeric, like I-70 instead of interstate 70 or something like that. I have not added this logic yet, just have the road_number field for improvement in the future. The later code only adds the road_number if it is different from the road_name

wzdx/standard_to_enhanced/planned_events_translator.py
jacob6838: This was just for testing, removing now

wzdx/tools/polygon_tools.py
jacob6838: yeah, it does its best but can easily fail. It is just for worst case scenarios, in iCone messages where the direction is not specified, and I need to choose one

wzdx/raw_to_standard/planned_events.py
```diff
@@ -202,6 +203,7 @@ def hex_to_binary(hex_string):
     "left lane": 'general',
     "center lane": 'general',
     "middle two lanes": 'general',
+    'general': 'general',
```
payneBrandon:
```suggestion
    "general": 'general',
```
Just to keep consistent

payneBrandon (changes requested):
A few of the tests are currently failing:

`================================================================================== short test summary info ================================================================================== 
FAILED tests/raw_to_standard/planned_events_test.py::test_generate_standard_messages_from_string - AssertionError: assert [{'event': {'...': 'we234de'}] == [{'event': {'...: '23wsg54h'}]    
FAILED tests/raw_to_standard/planned_events_test.py::test_get_lanes_list_1 - AssertionError: assert [{'order': 1,...': 'general'}] == [{'order': 1,...: 'shoulder'}]
FAILED tests/standard_to_enhanced/planned_events_translator_test.py::test_parse_work_zone_linestring - AssertionError: assert {'geometry': ...e': 'Feature'} == {'geometry': ...e': 'Feature'}FAILED tests/standard_to_enhanced/planned_events_translator_test.py::test_wzdx_creator - AssertionError: assert {'features': ...reCollection'} == {'features': ...reCollection'}`

PR #42, opened 2022-05-18
Opentms incidents

jacob6838:
Adding OpenTMS-Incident messages to the WZDx translator

- Tested in RTDH Test
- Added is_incident method to determine whether a msg is an incident and whether it has the correct metadata to be a work zone. Integrated into message translation to only process work zone incidents
- Added create_description method to create new description when the 511 CARS error is present. Not rigorously defined, just thrown together, this was a low priority
- Added and updated unit tests


jacob6838:
@payneBrandon This is a fairly minor update to the planned events translator. I have tested the change in RTDH Test, and Tony wants me to get my latest changes into Main so that we can publish some of the code/documentation

README.md
```diff
@@ -9,7 +9,10 @@ Work zone code and documentation for WZDx, iCone, etc.
...
+This is an open source, proof of concept solution for translating work zone data in the form of CDOT Planned Events to the standardized WZDx 4.0 format, as well as having additional translators to translate COtrip/Salesforce, iCone, and NavJOY messages to the WZDx 3.1 format. This project was developed for CDOT. A unique translator has been developed for each of these message types. These translators read in the source message, parse out specific fields, and generate a WZDx message. For more information on these message formats and the data mappings between these messages and the WZDx format, see the [documentation](wzdx/docs). sample_files are located [here](wzdx/sample_files). All these translators are built to run from the command line and from GCP Dataflows, hosted within the CDOT OIM WZDX environment, connected to the RTDH (real time data hub). For more information on cloud hosting, see [GCP_cloud_function](wzdx/GCP_cloud_function).
```
payneBrandon: This states 3.1 format, but aren't we now targeting 4?

wzdx/raw_to_standard/planned_events.py
```diff
@@ -287,13 +318,27 @@ def all_lanes_open(lanes):
...
-        logging.warning(
-            f"Invalid message with id: '{pd.get('properties/id')}'' because description matches invalid event description: '{INVALID_EVENT_DESCRIPTION}'")
-        return {}
+# On {roadName}, between mile markers {startMarker} and {endMarker}. {typeOfWork}. Running between {startTime} and {endTime}
+def create_description(name, roadName, startMarker, endMarker, typeOfWork, startTime, endTime):
+    return f"Event {name}, on {roadName}, between mile markers {startMarker} and {endMarker}. {typeOfWork}. Running between {startTime} and {endTime}"
+
+
+def create_rtdh_standard_msg(pd, isIncident):
+    description = pd.get('properties/travelerInformationMessage')
```
payneBrandon, line 322: Do we want to plug this description into the return message somewhere? It looks like we create it but don't use it anywhere yet. Maybe that's to come?

wzdx/raw_to_standard/planned_events.py
```diff
@@ -312,6 +357,9 @@ def create_rtdh_standard_msg(pd):
                         date_tools.parse_datetime_from_iso_string)
     end_date = pd.get("properties/clearTime",
                       date_tools.parse_datetime_from_iso_string)
+    if not end_date:
+        # Since there is no end date, assume still active, set end date in future
+        end_date = datetime.datetime.utcnow() + datetime.timedelta(hours=24)
```
payneBrandon: Since we're looking at planned events here, the start date could be in the future as well right? Should we instead add 24 hours to the start date so we don't end up with end date before start date for edge cases?

wzdx/raw_to_standard/planned_events.py
```diff
@@ -154,6 +180,7 @@ def hex_to_binary(hex_string):
 # TODO: Consider support road closures
 DEFAULT_EVENT_TYPE = ('work-zone', [])
 EVENT_TYPE_MAPPING = {
```
payneBrandon, line 176: I like this mapping technique you've got going on here, thinking I'll somehow steal it for the WZDx to TIM mapping app.

wzdx/raw_to_standard/planned_events.py
```diff
@@ -64,15 +73,25 @@ def generate_standard_messages_from_string(input_file_contents):
     return standard_messages
 
 
+# TODO: Integrate Category
+def is_incident(msg):
```
payneBrandon: Since we're returning two values, maybe rename "is_incident_workzone" or something to show we're getting both?

wzdx/raw_to_standard/planned_events.py
```diff
@@ -287,13 +318,27 @@ def all_lanes_open(lanes):
...
-    if pd.get('properties/travelerInformationMessage') == INVALID_EVENT_DESCRIPTION:
-        logging.warning(
-            f"Invalid message with id: '{pd.get('properties/id')}'' because description matches invalid event description: '{INVALID_EVENT_DESCRIPTION}'")
-        return {}
+# On {roadName}, between mile markers {startMarker} and {endMarker}. {typeOfWork}. Running between {startTime} and {endTime}
+def create_description(name, roadName, startMarker, endMarker, typeOfWork, startTime, endTime):
+    return f"Event {name}, on {roadName}, between mile markers {startMarker} and {endMarker}. {typeOfWork}. Running between {startTime} and {endTime}"
+
+
+def create_rtdh_standard_msg(pd, isIncident):
```
payneBrandon, line 321: I assume we added the `isIncident` for later use?

wzdx/raw_to_standard/planned_events.py
```diff
@@ -64,15 +73,25 @@ def generate_standard_messages_from_string(input_file_contents):
...
+    # is_wz = WORK_ZONE_INCIDENT_TYPES.get(type, {}).get(category)
+    is_wz = WORK_ZONE_INCIDENT_TYPES.get(type, {}) != {}
+    return is_incident, is_wz
+
+
 def generate_raw_messages(message_string):
     msg = json.loads(message_string)
     messages = []
 
     separated_messages = expand_event_directions(msg)
```
payneBrandon: Are we planning to add validation/checks for this below? Looks like we removed the call to `validate_closure`. If not, we could just `return expand_event_directions(msg)` and forgo the loop.

wzdx/raw_to_standard/planned_events.py
```diff
@@ -18,16 +20,23 @@
...
+    "Other":                    {"Environmental": True},
+    "Emergency Roadwork":       {"Traffic": True},
+}
+INCIDENT_ID_REGEX = "^OpenTMS-Incident"
+
 
 def main():
-    navjoy_file, output_dir = parse_rtdh_arguments()
-    input_file_contents = open(navjoy_file, 'r').read()
+    source_file, output_dir = parse_rtdh_arguments()
```
payneBrandon, line 32: Good catch!

tests/standard_to_enhanced/planned_events_translator_test.py
```diff
@@ -362,7 +361,6 @@ def test_wzdx_creator(mock_dt, mock_dt_3, _):
                         "direction": "westbound",
                         "relationship": {},
                         "description": "Between Exit 49: CO 65; Grand Mesa (5 miles east of the Palisade area) and US 6 (Debeque) from Mile Point 50 to Mile Point 60. Road closed expect delays due to bridge construction. Until May 1, 2022 at about 12:26PM MDT.",
-                        'creation_date': "2021-10-29T18:26:04Z",
```
payneBrandon, line 365: Are we not parsing the creation_date value anymore?

tests/raw_to_standard/planned_events_test.py
```diff
@@ -215,3 +207,68 @@ def test_get_lanes_list_2():
...
+            "id": "OpenTMS-Incident2028603626",
+            "type": "Chain Law Code 18"
+        }
+    }
+    expected = (True, False)
+    actual = planned_events.is_incident(msg)
+    assert actual == expected
+
+
+def test_is_incident_false_false():
```
payneBrandon, line 237: Do we need a "_false_true" test in here to shore up?

payneBrandon (changes requested):
Looking good!

wzdx/raw_to_standard/planned_events.py
jacob6838: This is just to cover the case when the end date is null but the start date is valid

payneBrandon: Right, but could we end up with a start date of say next Tuesday since this is a planned event? Then adding end date of tomorrow doesn't make sense.

wzdx/raw_to_standard/planned_events.py
jacob6838: yeah, I thought I would need it but there is honestly nothing different for them. There might be eventually though, so I want to leave it here

wzdx/raw_to_standard/planned_events.py
jacob6838: I removed it because everything it handles was already handled by the schema validation (from confluent) completed in the GCP dataflow. Didn't really think about uses other than GCP, can add it back if you want
Removed loop

wzdx/raw_to_standard/planned_events.py
jacob6838: copy past errors lol

tests/raw_to_standard/planned_events_test.py
jacob6838: no, if it is a planned event, then the incident work zone type check doesn't make any sense anymore. All of the types for planned events are handled by the EVENT_TYPE_MAPPING dict

tests/standard_to_enhanced/planned_events_translator_test.py
jacob6838: correct. It was not actually present in the planned events messages, there is just an update date. It is an optional field, so I am removing it

wzdx/raw_to_standard/planned_events.py
jacob6838: renamed to is_incident_wz

wzdx/raw_to_standard/planned_events.py
jacob6838: overusing dictionaries is kinda my specialty

wzdx/raw_to_standard/planned_events.py
jacob6838: Good catch, I referenced this in the message now

README.md
jacob6838: Updated

payneBrandon (approved):

PR #56, opened 2024-11-08
Updating Message Samples + MultiPoint

jacob6838:
- Updating message samples for planned events and icone datasets
- Using geometry type "MultiPoint" for events without enhanced geometry


payneBrandon (commented):
Looks like we have two "production_sample.json" files, one under /enhanced and one under /enhanced/planned_events. They appear to be identical, do we need both?

payneBrandon:
What are the files under `experimental/prod_experimental_diff_20230522-140004/` used for?

jacob6838:
> What are the files under `experimental/prod_experimental_diff_20230522-140004/` used for?

Those are an example a planned event being modified by the experimental_combination/icone script. The diff file shows exactly what was modified by the combination script, and the others are raw values for each input/output. I can add a README here to describe it

payneBrandon:
> > What are the files under `experimental/prod_experimental_diff_20230522-140004/` used for?
> 
> Those are an example a planned event being modified by the experimental_combination/icone script. The diff file shows exactly what was modified by the combination script, and the others are raw values for each input/output. I can add a README here to describe it

Ah, yeah let's get a README in there, this is definitely useful addition

jacob6838:
> > > What are the files under `experimental/prod_experimental_diff_20230522-140004/` used for?
> > 
> > 
> > Those are an example a planned event being modified by the experimental_combination/icone script. The diff file shows exactly what was modified by the combination script, and the others are raw values for each input/output. I can add a README here to describe it
> 
> Ah, yeah let's get a README in there, this is definitely useful addition

Just pushed this new README!

wzdx/raw_to_standard/planned_events.py
```diff
@@ -683,14 +683,32 @@ def get_cross_streets_from_description(description: str) -> tuple[str, str]:
         return ("", "")
 
 
+def get_mileposts_from_description(message: str) -> tuple[str, str]:
```
payneBrandon: This is a really interesting function, let's add some unit tests to show capabilities and limitations so we know what to expect when the throw us a curve ball down the line

payneBrandon: Also, is this just making the assumption that the Exit number will be the same as milepost?

wzdx/raw_to_standard/planned_events.py
```diff
@@ -683,14 +683,32 @@ def get_cross_streets_from_description(description: str) -> tuple[str, str]:
...
+        return m.group(1, 2)
+    except:
+        return ("", "")
+
+
 def get_route_details_for_coordinates_lngLat(
-    cdotGeospatialApi: cdot_geospatial_api.GeospatialApi, coordinates: list[list[float]]
+    cdotGeospatialApi: cdot_geospatial_api.GeospatialApi,
+    coordinates: list[list[float]],
+    reversed: bool,
```
payneBrandon: Doesn't look like we use this bool, do we need it?

payneBrandon (changes requested):
Just a couple questions

wzdx/raw_to_standard/planned_events.py
jacob6838: The description text is currently automatically generated, so for now, it parses well. Good point that I was processing the exit numbers, I switched this to read the Mile Post measurements. I also implemented the following priorities for determine mileposts:
1. geospatial route details, if start and end routes match
2. planned event properties (startMarker and endMarker)
3. description parsing

wzdx/raw_to_standard/planned_events.py
jacob6838: removed!

payneBrandon (approved):
lgtm! Tests pass and comments have been addressed

mwodahl (approved):
This all looks good to me!

